PR #88, opened 2023-03-04
small bug fixed in tests library page

Ark2307:


apps/dashboard/web/src/apps/dashboard/views/marketplace/PageMarketplace.vue
```diff
@@ -168,6 +168,9 @@ export default {
         }
     },
     methods: {
+        logError(err){
```
avneesh-akto, line 171: remove this method plz

PR #190, opened 2023-03-07
Solved issue number 120 of hackfest

Ark2307:


ankush-jain-akto:
Do this the new way 😃 
We should pass `$event` all the way up and let the caller decide about opening in a new tab.

PR #193, opened 2023-03-09
Added changes related to mass assignment attack

aktoboy:


apps/testing/src/main/java/com/akto/rules/CreateAdminUserViaMassAssignment.java
```diff
@@ -0,0 +1,180 @@
...
+
+import java.util.*;
+
+public class CreateAdminUserViaMassAssignment extends TestPlugin{
+
+    private final String testRunId;
+    private final String testRunResultSummaryId;
+    Logger logger = LoggerFactory.getLogger(CreateAdminUserViaMassAssignment.class);
+    private static final Set<String> allowedMethods = new HashSet<>(Arrays.asList("POST", "PUT", "PATCH"));
+    private static final List<String> allowedEndpointNames = Arrays.asList("users", "Users", "user", "User");
```
ayushaga14, line 27: can be done via regex as well instead of naming all combinations

apps/testing/src/main/java/com/akto/rules/CreateAdminUserViaMassAssignment.java
```diff
@@ -0,0 +1,180 @@
...
+        BasicDBObject flattenedResponse = JSONUtils.flattenWithDots(RequestTemplate.parseRequestPayload(response.getJsonResponseBody(), null));
+        if(flattenedRequest.keySet().size() > flattenedResponse.keySet().size()){
+            throw new Exception("Request body has more params than response body");
+        }
+        Set<String> extraParams = new HashSet<>();
+        for(String responseKey: flattenedResponse.keySet()){
+            String[] splitResponseKeys = responseKey.split("\\.");
+            for(String splitResponseKey: splitResponseKeys){
+                for(String requestKey: flattenedRequest.keySet()){
+                    if(!requestKey.contains(splitResponseKey)){
```
ayushaga14, line 162: might not work properly in nested payloads, since here we are splitting the key. Only keys at root level will be detected. Do check once

apps/testing/src/main/java/com/akto/rules/CreateAdminUserViaMassAssignment.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    private String createBodyForTest(OriginalHttpRequest request, OriginalHttpResponse response, Set<String> extraParams) {
+        String jsonRequestBody = request.getJsonRequestBody();
+        BasicDBObject jsonRequestBodyObj = BasicDBObject.parse(jsonRequestBody);
+        //Update the email id to a random one
+        for(String key: allowedKeysInRequest){
+            if(jsonRequestBodyObj.containsField(key)){
+                String email = jsonRequestBodyObj.getString(key);
+                jsonRequestBodyObj.put(key, UUID.randomUUID() + "@" + email.split("@")[1]);
```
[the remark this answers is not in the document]
aktoboy: I think this wont be an issue, ref [this](https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address)

apps/testing/src/main/java/com/akto/rules/CreateAdminUserViaMassAssignment.java
aktoboy: Lets do this via keywords for now, the current implementation doesn't cover all the cases, lets learn more from customers and later we will make this test even more comprehensive.

apps/testing/src/main/java/com/akto/rules/CreateAdminUserViaMassAssignment.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+    }
+
+    private String createBodyForTest(OriginalHttpRequest request, OriginalHttpResponse response, Set<String> extraParams) {
+        String jsonRequestBody = request.getJsonRequestBody();
+        BasicDBObject jsonRequestBodyObj = BasicDBObject.parse(jsonRequestBody);
+        //Update the email id to a random one
+        for(String key: allowedKeysInRequest){
+            if(jsonRequestBodyObj.containsField(key)){
```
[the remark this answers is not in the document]
aktoboy: This was an intentional call, current implementation will try to replace email only at root level. We will make this more robust in a later iteration.

apps/testing/src/main/java/com/akto/rules/CreateAdminUserViaMassAssignment.java
aktoboy: requestKey contains a dot seperated flattened key and splitResponseKey contains the actual leaf level key from response and since this a .contains check I dont think we will miss any keys.

PR #202, opened 2023-03-16
hit api if no response in postman.json

avneesh-akto:


apps/dashboard/src/main/java/com/akto/utils/Utils.java
```diff
@@ -39,63 +48,107 @@ public static Map<String, String> getVariableMap(ArrayNode variables){
...
+            sb.append(val);
+        }
+
+        matcher.appendTail(sb);
+
+        return sb.toString();
+    }
     
-    public static Map<String, String> convertApiInAktoFormat(JsonNode apiInfo, Map<String, String> variables, String accountId) {
+    public static Map<String, String> convertApiInAktoFormat(JsonNode apiInfo, Map<String, String> variables, String accountId, boolean allowReplay) {
```
ankush-jain-akto, line 79: Can we add some tests for this?

PR #203, opened 2023-03-17
Feature/consistent tables

ankush-jain-akto:


aktoboy (approved):

Ark2307 (approved):

Ark2307 (commented):
pagination is not updated
active row color is not same
width of headers is in same row
actions not visible
csv button is not updated

PR #244, opened 2023-03-31
fixed method value of

avneesh-akto:


apps/api-runtime/src/main/java/com/akto/runtime/APICatalogSync.java
```diff
@@ -1206,89 +1206,98 @@ public void buildFromDB(boolean calcDiff, boolean fetchAllSTI) {
         }
     }
 
+    private static void build(SingleTypeInfo param, Map<Integer, APICatalog> ret) {
```
ankush-jain-akto, line 1209: rename the function to `buildHelper` please ?

avneesh-akto: done

PR #251, opened 2023-04-03
Update fintech.json

bhavya-error404:
Added regex for google ads cookies.


ankush-jain-akto:
Hi @bhavya-error404 - what is the source for this regex? Or if you can point us to the right website/page where google-cookie information is mentioned. That would be great!

bhavya-error404:
> Hi @bhavya-error404 - what is the source for this regex? Or if you can point us to the right website/page where google-cookie information is mentioned. That would be great!

bhavya-error404:
Hi Ankit!

I have used https://business.safety.google/adscookies/ as a reference for google ads cookie type.

PR #311, opened 2023-05-20
try noauth only if header is present

ayushaga14:


apps/testing/src/main/java/com/akto/test_editor/execution/Executor.java
```diff
@@ -259,7 +259,10 @@ public ExecutorSingleOperationResp runOperation(String operationType, RawApi raw
             case "remove_auth_header":
                 List<String> authHeaders = (List<String>) varMap.get("auth_headers");
                 for (String header: authHeaders) {
-                    Operations.deleteHeader(rawApi, header);
+                    ExecutorSingleOperationResp resp = Operations.deleteHeader(rawApi, header);
+                    if (resp.getErrMsg().contains("header key not present")) {
```
ankush-jain-akto, line 263: We really shouldn't be dependent on error messages for our logic-flows. Not for this release, but please make sure you fix this later next week.

PR #322, opened 2023-05-30
Feature/test editor files versioning isactive

oren-akto:


apps/dashboard/src/main/java/com/akto/utils/DashboardVersion.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        
+        return 0;
+    }
+
+    public static String getDashboardVersion() {
+        AccountSettings accountSettings = AccountSettingsDao.instance.findOne(new BasicDBObject());
+        String dashboardVersionDb = accountSettings.getDashboardVersion();
+
+        String aktoImageTag = dashboardVersionDb.split(" - ", 2)[0];
```
ayushaga14, line 64: handle null check here, in case dashboardVersionDb is null, or array size is less than 3 after split

apps/dashboard/src/main/java/com/akto/listener/InitializerListener.java
```diff
@@ -1053,14 +1058,36 @@ public static void updateTestEditorTemplatesFromGithub() {
...
+                        templateMinimumAktoVersion = metadata.getMinAktoVersion();
+                    } else {
+                        templateMinimumAktoVersion = metadata.getMinOnpremVersion(); 
+                    }
+
+                    // Get dashboard version
+                    String dashboardVersion = DashboardVersion.getDashboardVersion();
+
+                    // Check if updated template is not supported by dashboard
+                    if (DashboardVersion.isSemanticVersionString(dashboardVersion)) {
```
ayushaga14, line 1079: so basically if tag is not semantic(local/stage deployments), then we will always update, right? since if check will get skipped

oren-akto: Yes

apps/dashboard/src/main/java/com/akto/listener/InitializerListener.java
```diff
@@ -1053,14 +1062,36 @@ public static void updateTestEditorTemplatesFromGithub() {
...
+                        if (DashboardVersion.compareVersions(templateMinimumAktoVersion, dashboardVersion) > 0) {
```
ayushaga14, line 1084: 0 is being returned if either of the version isn't semantic, is this correct?

PR #328, opened 2023-06-05
Feature/akto saas 2

notshivansh:


apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -95,88 +103,91 @@ public static void main(String[] args) throws InterruptedException {
...
-            }
+                loggerMaker.infoAndAddToDb("Found one + " + testingRun.getId().toHexString(), LogDb.TESTING);
+
+                try {
+                    AccountSettings accountSettings = AccountSettingsDao.instance.findOne(new BasicDBObject());
+                    boolean runStatusCodeAnalyser = accountSettings == null ||
+                            accountSettings.getSetupType() != AccountSettings.SetupType.PROD;
+
+                    SampleMessageStore sampleMessageStore = SampleMessageStore.create();
+                    AuthMechanismStore authMechanismStore = AuthMechanismStore.create();
```
ayushaga14, line 142: Not sure if we need to bring sample message fetch etc here. Previous code should work imo, but we can check once

ayushaga14: Also main file shouldn't have a lot of logic, gets harder to maintain later on

apps/testing/src/main/java/com/akto/testing/TestExecutor.java
```diff
@@ -95,24 +102,23 @@ public void workflowInit (TestingRun testingRun, ObjectId summaryId) {
...
         int now = Context.now();
         int maxConcurrentRequests = testingRun.getMaxConcurrentRequests() > 0 ? testingRun.getMaxConcurrentRequests() : 100;
         TestingEndpoints testingEndpoints = testingRun.getTestingEndpoints();
 
-        Map<String, SingleTypeInfo> singleTypeInfoMap = SampleMessageStore.buildSingleTypeInfoMap(testingEndpoints);
-        Map<ApiInfo.ApiInfoKey, List<String>> sampleMessages = SampleMessageStore.fetchSampleMessages();
-        List<TestRoles> testRoles = SampleMessageStore.fetchTestRoles();
-        AuthMechanism authMechanism = AuthMechanismsDao.instance.findOne(new BasicDBObject());
+        sampleMessageStore.buildSingleTypeInfoMap(testingEndpoints);
+        List<TestRoles> testRoles = sampleMessageStore.fetchTestRoles();
```
ayushaga14, line 112: same here, i think old code should work as it is ideally

apps/testing/src/main/java/com/akto/store/SampleMessageStore.java
```diff
@@ -62,21 +63,33 @@ public static Map<String, SingleTypeInfo> buildSingleTypeInfoMap(TestingEndpoint
...
+        SampleMessageStore ret = new SampleMessageStore();
+        ret.fetchSampleMessages();
+        return ret;
+    }
 
-        return singleTypeInfoMap;
+    public static SampleMessageStore createWithSampleDataMap(Map<ApiInfo.ApiInfoKey, List<String>> sampleDataMap) {
+        SampleMessageStore ret = new SampleMessageStore();
+        ret.sampleDataMap = sampleDataMap;
+        return ret;
```
ayushaga14, line 84: let's not have 2 different methods just because o new argument

apps/testing/src/main/java/com/akto/store/SampleMessageStore.java
```diff
@@ -27,9 +27,10 @@ public class SampleMessageStore {
 
 
     private static final LoggerMaker loggerMaker = new LoggerMaker(SampleMessageStore.class);
-    public static Map<String, SingleTypeInfo> buildSingleTypeInfoMap(TestingEndpoints testingEndpoints) {
-        Map<String, SingleTypeInfo> singleTypeInfoMap = new HashMap<>();
-        if (testingEndpoints == null) return singleTypeInfoMap;
+    private Map<ApiInfo.ApiInfoKey, List<String>> sampleDataMap = new HashMap<>();
+    private Map<String, SingleTypeInfo> singleTypeInfos = new HashMap<>();
```
ayushaga14, line 31: also check if these changes are required once, defining these new class variables

apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -95,88 +103,91 @@ public static void main(String[] args) throws InterruptedException {
             }
         } while (!connectedToMongo);
 
-        int delta = Context.now() - 20*60;
         setupRateLimitWatcher();
```
ayushaga14, line 106: how are we handling rate limit per account?

apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -60,30 +66,32 @@ private static ObjectId createTRRSummaryIfAbsent(TestingRun testingRun, int star
...
-                Context.accountId.set(1_000_000);
-                AccountSettings settings = AccountSettingsDao.instance.findOne(AccountSettingsDao.generateFilter());
-                if (settings == null) {
-                    return;
-                }
-                int globalRateLimit = settings.getGlobalRateLimit();
-                Map<ApiRateLimit, Integer> rateLimitMap =  RateLimitHandler.getInstance().getRateLimitsMap();
-                rateLimitMap.clear();
-                rateLimitMap.put(new GlobalApiRateLimit(globalRateLimit), globalRateLimit);
+                AccountTask.instance.executeTask(new Consumer<Account>() {
```
ayushaga14, line 70: didn't get the working of this block of code. So basically this will iterate on all active accounts, and override  rateLimitMap

PR #339, opened 2023-06-12
Sql injection support

ayushaga14:


ankush-jain-akto:
To be released after m x n

PR #356, opened 2023-06-19
added 3 lfi tests

avneesh-akto:


apps/dashboard/src/main/resources/inbuilt_test_yaml_files/LFIInPath.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+execute:
+  type: single
+  requests:
+    - req:
+      - modify_url:
+          regex_replace:
+            regex: \/$
+            replace_with: ''
+      - modify_url: ${urlVar}/${filePaths}
```
ayushaga14, line 71: this won't work currently with develop changes, since filepaths is a list, remove this test for now.

PR #358, opened 2023-06-19
added 3 open redirect tests

avneesh-akto:


apps/dashboard/src/main/resources/inbuilt_test_yaml_files/OpenRedirectSubdomainWhitelist.yaml
```diff
@@ -0,0 +1,77 @@
...
+            extract: payload_location_value
+          key:
+            extract: payload_location_key
+
+execute:
+  type: single
+  requests:
+    - req:
+      - modify_query_param:
+          payload_location_key: http://${payload_location_value}.evil.com/
```
ayushaga14, line 60: will appending evil.com work?

apps/dashboard/src/main/resources/inbuilt_test_yaml_files/OpenRedirectSubdomainWhitelist.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - "https://portswigger.net/kb/issues/00500100_open-redirection-reflected"
+api_selection_filters:
+  response_code:
+    eq: 302
+  response_headers:
+    for_one:
+      key: 
+        eq: location
+      value:
+        regex: .*
```
ayushaga14, line 30: regex condition is redundant, since key condition is already present

PR #362, opened 2023-06-20
Feature/misconfig tests

avneesh-akto:


apps/dashboard/src/main/resources/inbuilt_test_yaml_files/FirebaseUnauthenticated.yaml
```diff
@@ -0,0 +1,42 @@
...
+    lt: 300
+  url:
+    regex: .*
+    extract: urlVar
+
+execute:
+  type: single
+  requests:
+    - req:
+        - modify_url: ${urlVar}.json
```
ayushaga14, line 34: got to make sure ${urlVar} has '/' at the end

PR #387, opened 2023-06-27
Template optimizations and new features

ayushaga14:


libs/dao/src/main/java/com/akto/dao/test_editor/strategy/StrategyParser.java
```diff
@@ -0,0 +1,25 @@
...
+    
+    public Strategy parse(Object metadataObj) {
+        Map<String, Object> metadataMap = (Map) metadataObj;
+        Strategy strategy = new Strategy();
+
+        if (metadataMap == null) {
+            return strategy;
+        }
+
+        Object val = metadataMap.get("run_once");
```
avneesh-akto, line 17: Will u do this for all the fields. Use gson or something so that if u define setter it will automatically parse

apps/testing/src/main/java/com/akto/test_editor/filter/FilterAction.java
```diff
@@ -168,7 +168,7 @@ public DataOperandsFilterResponse applyFilterOnUrl(FilterActionRequest filterAct
     }
 
     public void extractUrl(FilterActionRequest filterActionRequest, Map<String, Object> varMap) {
-        String url = filterActionRequest.getApiInfoKey().getUrl();
+        String url = filterActionRequest.getRawApi().getRequest().getUrl();
```
avneesh-akto, line 171: this was done to get api/books/1 instead of api/books/INTEGER right??

PR #409, opened 2023-07-04
added 12 config test templates

Rishav1919:


ankush-jain-akto:
merged in feature/auth0_integration already

PR #529, opened 2023-08-28
fix redundant data type updates

notshivansh:


libs/dao/src/main/java/com/akto/dto/data_types/Conditions.java
```diff
@@ -61,4 +61,14 @@ public String toString() {
             ", operator='" + getOperator() + "'" +
             "}";
     }
+
+    public static boolean areEqual(Conditions a, Conditions b) {
```
shivamrawat101192, line 65: @notshivansh add this method in Conditions class overriding equals method

notshivansh: @shivamrawat101192 added the same.

ankush-jain-akto (approved):

PR #576, opened 2023-09-19
Feature/login via GitHub

ankush-jain-akto:


apps/dashboard/src/main/java/com/akto/utils/GithubLogin.java
```diff
@@ -0,0 +1,54 @@
+package com.akto.utils;
+
+import com.akto.dao.ConfigsDao;
+import com.akto.dao.context.Context;
+import com.akto.dto.Config;
+import com.akto.dto.Config.GithubConfig;
+
+public class GithubLogin {
+
+    public static final int PROBE_PERIOD_IN_SECS = 60;
```
aktoboy, line 10: We can make this private

apps/dashboard/src/main/java/com/akto/action/user/GithubSsoAction.java
```diff
@@ -0,0 +1,89 @@
...
+            addActionError("Only admin can delete SSO");
+            return ERROR.toUpperCase();
+        }
+
+        DeleteResult result = ConfigsDao.instance.deleteAll(Filters.eq("_id", "GITHUB-ankush"));
+
+        if (result.getDeletedCount() > 0) {
+            for (Object obj : UsersDao.instance.getAllUsersInfoForTheAccount(Context.accountId.get())) {
+                BasicDBObject detailsObj = (BasicDBObject) obj;
+                UsersDao.instance.updateOne("login", detailsObj.getString(User.LOGIN), Updates.set("refreshTokens", new ArrayList<>()));
```
aktoboy, line 34: Shouldn't we unset GithubSignupInfo stored inside user document here?

apps/api-runtime/src/main/java/com/akto/runtime/APICatalogSync.java
```diff
@@ -1383,17 +1376,23 @@ public void syncWithDB(boolean syncImmediately, boolean fetchAllSTI) {
...
-
-                loggerMaker.infoAndAddToDb((System.currentTimeMillis() - start) + ": " + res.getInserts().size() + " " +res.getUpserts().size(), LogDb.RUNTIME);
+            do {
+
+                List<WriteModel<SingleTypeInfo>> slicedWrites = writesForParams.subList(from, Math.min(from + batch, writesForParams.size()));
+                from += batch;
+                BulkWriteResult res =
+                        SingleTypeInfoDao.instance.getMCollection().bulkWrite(
+                                slicedWrites,
+                                new BulkWriteOptions().ordered(true).bypassDocumentValidation(false)
```
aktoboy, line 1391: Ordered should be set to false, else even if a single write fails, the entire batch will fail

aktoboy (changes requested):
Few comments, rest lgtm.

PR #599, opened 2023-09-29
Feature/test editor files GitHub sync

oren-akto:
Requires - https://github.com/akto-api-security/tests-library/pull/16


apps/dashboard/src/main/java/com/akto/action/AccountAction.java
```diff
@@ -261,9 +261,9 @@ public void run() {
                 RuntimeListener.addSampleData();
                 AccountSettingsDao.instance.updateOnboardingFlag(true);
                 InitializerListener.insertPiiSources();
-                InitializerListener.saveTestEditorYaml();
                 try {
                     InitializerListener.executePIISourceFetch();
+                    InitializerListener.updateTestEditorTemplatesFromGithub(newAccountId);
```
ayushaga14, line 266: what if pii source fetch fails? updateTestEditorTemplatesFromGithub will get skipped

PR #601, opened 2023-09-29
Avoid re-running tests

aktoboy:


apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -149,6 +152,26 @@ public static void main(String[] args) throws InterruptedException {
...
+                                return;
+                            } else {
+                                loggerMaker.infoAndAddToDb("Test run was executed long ago, TRR_ID:"
+                                        + testingRunResult.getHexId() + ", TRRS_ID:" + testingRunResultSummary.getHexId() + " TR_ID:" + testingRun.getHexId(), LogDb.TESTING);
+                                TestingRunResultSummariesDao.instance.updateOne(Filters.eq(TestingRunResultSummary.ID, testingRunResultSummary.getId()), Updates.set(TestingRunResultSummary.STATE, TestingRun.State.FAILED));
+                            }
+                        } else {
+                            loggerMaker.infoAndAddToDb("No executions made for this test, will need to restart it, TRRS_ID:" + testingRunResultSummary.getHexId() + " TR_ID:" + testingRun.getHexId(), LogDb.TESTING);
+                            TestingRunResultSummariesDao.instance.updateOne(Filters.eq(TestingRunResultSummary.ID, testingRunResultSummary.getId()), Updates.set(TestingRunResultSummary.STATE, TestingRun.State.FAILED));
```
ayushaga14, line 172: can get stuck in loop if due to some reason test run results are empty for a test

aktoboy: As discussed this wont get stuck because we are picking tests after 20 minutes which are in running state

ayushaga14 (approved):

PR #613, opened 2023-10-03
OpenAPI Misconfiguration: Missing Global Security Requirement (Enviro…

xShuden:
Global security requirements inform API consumers of the authentication and authorization parameters required to successfully interact with the API.

An absent global security definition might enable attackers to interact with sensitive API endpoints and allow them to perform operations that should be restricted to specific user accounts with specific privileges.


avneesh-akto:
This OpenApi file is not being used anywhere. Closing PR.

PR #616, opened 2023-10-04
fix for multiple zip archive downloads

oren-akto:


apps/dashboard/src/main/java/com/akto/listener/InitializerListener.java
```diff
@@ -1238,14 +1238,18 @@ private static String getUpdateDeploymentStatusUrl() {
...
         scheduler.scheduleAtFixedRate(new Runnable() {
             public void run() {
                 AccountTask.instance.executeTask(new Consumer<Account>() {
                     @Override
                     public void accept(Account t) {
                         try {
                             int accountId = t.getId();
-                            updateTestEditorTemplatesFromGithub(accountId);
+                            loggerMaker.infoAndAddToDb(String.format("Updating akto test templates for account: %d", accountId), LogDb.DASHBOARD);
+                            processTemplateFilesZip(repoZip);
```
ayushaga14, line 1252: null check

ayushaga14 (approved):

PR #618, opened 2023-10-05
add error messages for failing test results

notshivansh:


apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -196,6 +197,42 @@ public static void main(String[] args) throws InterruptedException {
                         Filters.eq("_id", testingRun.getId()),  completedUpdate
                 );
 
+                if(summaryId != null){
+                    TestingRunResultSummary testingRunResultSummary = TestingRunResultSummariesDao.instance.findOne(
+                        Filters.eq(TestingRunResultSummary.ID, summaryId));
+                    if(testingRunResultSummary.getState() == TestingRun.State.RUNNING
+                            || testingRunResultSummary.getState() == TestingRun.State.SCHEDULED){
+
```
ayushaga14, line 205: why are we recalculating all this(COUNT_ISSUES etc), if it's already being done inside apiwiseinit

apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -196,6 +197,42 @@ public static void main(String[] args) throws InterruptedException {
...
+                            totalApis = testingRun.getTestingEndpoints().returnApis().size();
+                        } catch (Exception e) {
+                            totalApis = 0;
+                        }
+
+                        Bson updates = Updates.combine(
+                                Updates.set(TestingRunResultSummary.END_TIMESTAMP, Context.now()),
+                                Updates.set(TestingRunResultSummary.STATE, State.COMPLETED),
+                                Updates.set(TestingRunResultSummary.COUNT_ISSUES, totalCountIssues),
+                                Updates.set(TestingRunResultSummary.TOTAL_APIS, totalApis),
```
ayushaga14, line 227: what's the use of total_apis

apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -196,6 +197,42 @@ public static void main(String[] args) throws InterruptedException {
                         Filters.eq("_id", testingRun.getId()),  completedUpdate
                 );
 
+                if(summaryId != null){
+                    TestingRunResultSummary testingRunResultSummary = TestingRunResultSummariesDao.instance.findOne(
+                        Filters.eq(TestingRunResultSummary.ID, summaryId));
+                    if(testingRunResultSummary.getState() == TestingRun.State.RUNNING
```
ayushaga14, line 203: why on scheduled?

apps/testing/src/main/java/com/akto/testing/Main.java
ayushaga14: let's not do this in this way. Let's identify what are the possible reasons why status is not getting updated and handle in properly. No additions in main.java

ayushaga14 (changes requested):
check comments

apps/testing/src/main/java/com/akto/testing/Main.java
notshivansh: Sure, let me refractor this PR to only add error messages for failing test runs.

ayushaga14 (approved):

PR #665, opened 2023-10-14
Update README.md

HarshMN2345:
made it easy to understand


avneesh-akto:
Hi @HarshMN2345 ,

The current Readme seems fine as it is. If you have specific suggestions or concerns, please let me know, and we can discuss further. Thanks!

PR #667, opened 2023-10-16
add unit tests for cicd and scheduled tests and add more logs

notshivansh:


apps/testing/src/main/java/com/akto/testing/ApiExecutor.java
```diff
@@ -25,7 +25,12 @@ private static OriginalHttpResponse common(Request request, boolean followRedire
 
         Integer accountId = Context.accountId.get();
         if (accountId != null) {
+            boolean rateLimitHit = true;
             while (RateLimitHandler.getInstance(accountId).shouldWait(request)) {
+                if(rateLimitHit){
+                    loggerMaker.infoAndAddToDb("Rate limit hit, sleeping", LogDb.TESTING);
```
ayushaga14, line 31: add timestamp to this log

ayushaga14 (commented):
minor comment

apps/testing/src/main/java/com/akto/testing/ApiExecutor.java
notshivansh: The loggerMaker adds a timestamp field to all logs in the DB by default, thus not needed here.

PR #694, opened 2023-10-26
remove filling testing results array an d extract summary update outside

notshivansh:


apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -197,6 +200,61 @@ public static void main(String[] args) throws InterruptedException {
                         Filters.eq("_id", testingRun.getId()),  completedUpdate
                 );
 
+                if(summaryId != null && testingRun.getTestIdConfig() != 1){
```
ayushaga14, line 203: do not keep whole logic in main.java, extract it out in a function and call that method here

PR #724, opened 2023-11-17
Feature/using collection ids

notshivansh:
Using collection IDs array for queries to add support for API groups.


libs/dao/src/main/java/com/akto/dto/ApiCollection.java
```diff
@@ -25,6 +31,16 @@ public class ApiCollection {
...
 
+    public enum Type {
+        API_GROUP
+    }
+
+    Type type;
+    public static final String _TYPE = "type";
+
+    List<CollectionCondition> conditions;
+    public static final String CONDITIONS_STRING = "conditions";
```
ankush-jain-akto, line 42: What is `CollectionCondition` and why we adding more members to `ApiCollection`?

libs/dao/src/main/java/com/akto/dto/ApiCollectionUsers.java
```diff
@@ -56,6 +68,108 @@ public enum CollectionType {
             });
         }});
 
+    public static Set<String> getUrlsFromConditions(List<CollectionCondition> conditions){
+        Set<ApiInfoKey> apis = new HashSet<>();
+        conditions.forEach((condition) -> {
+            if(CollectionCondition.Operator.AND.equals(condition.getOperator())){
+                apis.retainAll(condition.returnApis());
```
ankush-jain-akto, line 75: `apis` defaults to empty set. `retainAll` will be a no-op here

libs/dao/src/main/java/com/akto/dto/ApiCollectionUsers.java
```diff
@@ -56,6 +68,108 @@ public enum CollectionType {
...
+    }
+
+    public static void updateApiCollection(List<CollectionCondition> conditions, int id) {
+        Set<String> urls = getUrlsFromConditions(conditions);
+
+        ApiCollectionsDao.instance.updateOne(
+                Filters.eq(Constants.ID, id),
+                Updates.combine(
+                    Updates.set(ApiCollection.CONDITIONS_STRING, conditions),
+                    Updates.set(ApiCollection.URLS_STRING, urls)));
```
ankush-jain-akto, line 95: This should be `addToSet` ? `Updates.set` will override existing variable in db.

libs/dao/src/main/java/com/akto/dto/ApiCollectionUsers.java
```diff
@@ -56,6 +68,108 @@ public enum CollectionType {
...
+        updateCollectionsForCollectionId(filtersMap, update);
+    }
+
+    public static void addToCollectionsForCollectionId(List<CollectionCondition> conditions, int apiCollectionId) {
+        Bson update = Updates.addToSet(SingleTypeInfo._COLLECTION_IDS, apiCollectionId);
+        Bson matchFilter = Filters.nin(SingleTypeInfo._COLLECTION_IDS, apiCollectionId);
+        operationForCollectionId(conditions, apiCollectionId, update, matchFilter, false);
+    }
+
+    public static void removeFromCollectionsForCollectionId(List<CollectionCondition> conditions, int apiCollectionId) {
```
ankush-jain-akto, line 134: Can't follow much of this. These are all `public` methods?

libs/dao/src/main/java/com/akto/dto/CollectionConditions/ApiListCondition.java
```diff
@@ -0,0 +1,89 @@
...
+    public Map<CollectionType, Bson> returnFiltersMap(){
+        Map<CollectionType, Bson> filtersMap = new HashMap<>();
+        filtersMap.put(CollectionType.ApiCollectionId, new BasicDBObject());
+        filtersMap.put(CollectionType.Id_ApiCollectionId, new BasicDBObject());
+        filtersMap.put(CollectionType.Id_ApiInfoKey_ApiCollectionId, new BasicDBObject());
+
+        for(Map.Entry<CollectionType, Bson> filters: filtersMap.entrySet()){
+            List<Bson> apiFilters = new ArrayList<>();
+            CollectionType type = filters.getKey();
+            if(this.apiList != null && !this.apiList.isEmpty()){
```
ankush-jain-akto, line 77: This can be a huge query. Should be set carefully.

libs/dao/src/main/java/com/akto/dto/CollectionConditions/CollectionCondition.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Set;
+import org.bson.conversions.Bson;
+import com.akto.dto.ApiInfo;
+import com.akto.dto.ApiCollectionUsers.CollectionType;
+
+public abstract class CollectionCondition {
+    private Type type;
+    private Operator operator;
+
+    public CollectionCondition(Type type, Operator operator) {
```
ankush-jain-akto, line 13: There is a similar class already - `TestingEndpoints`

libs/dao/src/main/java/com/akto/dto/ApiCollection.java
notshivansh: collection condition is the filter which will be used to filter apis. We are adding more members to ApiCollection instead of creating another class, because, ApiCollections are used at some places

libs/dao/src/main/java/com/akto/dto/ApiCollectionUsers.java
notshivansh: You are correct. This is fixed in the followup PR. https://github.com/akto-api-security/akto/pull/740/files#diff-e8f80dd82161897bd4f7ea61b0d34f3ad0c3cf73cf8568ecebb054ef37fb3a85R84

libs/dao/src/main/java/com/akto/dto/ApiCollectionUsers.java
notshivansh: This function is used to add as well as remove APIs from a custom collection, thus using `Updates.set`

libs/dao/src/main/java/com/akto/dto/ApiCollectionUsers.java
notshivansh: Yes, apart from one or two methods, most are called outside.

libs/dao/src/main/java/com/akto/dto/CollectionConditions/ApiListCondition.java
notshivansh: I've a batching mechanism for updates of STI which updates in batches of 50,000. will use that for all updates.

libs/dao/src/main/java/com/akto/dto/CollectionConditions/CollectionCondition.java
notshivansh: sure, will inherit that class over this one to reduce code.

PR #752, opened 2023-12-05
batching error logs and update error messages

notshivansh:


libs/utils/src/main/java/com/akto/log/LoggerMaker.java
```diff
@@ -90,7 +120,27 @@ public void errorAndAddToDb(String err, LogDb db) {
         }
 
         if (db.equals(LogDb.BILLING) || db.equals(LogDb.DASHBOARD)) {
-            sendToSlack(err);
+            if (cache) {
+                try {
+                    if (errorLogCache.containsKey(err)) {
```
aktoboy, line 125: Replace this with getOrDefault

libs/utils/src/main/java/com/akto/log/CacheLoggerMaker.java
```diff
@@ -0,0 +1,60 @@
...
+
+        if (db.equals(LogDb.BILLING) || db.equals(LogDb.DASHBOARD)) {
+            try {
+                errorLogCache.put(err, errorLogCache.getOrDefault(err, 0) + 1);
+            } catch (Exception e) {
+                internalLogger.error("ERROR: In adding to log cache: " + e.toString());
+            }
+        }
+
+        if (errorLogCache.size() > CACHE_LIMIT) {
```
aktoboy, line 51: along with the size based condition, we should also add a time based condition. We will flush logs when either size breaches the CACHE_LIMIT or time condition is reached. We can keep time limit as 1 minute for now.

aktoboy: Without this logs might delayed till we reach limit of 50

PR #784, opened 2023-12-21
Feature/feature flag extended

notshivansh:


libs/utils/src/main/java/com/akto/mixpanel/AktoMixpanel.java
```diff
@@ -13,6 +15,8 @@
 
 public class AktoMixpanel {
     private static final Logger logger = LoggerFactory.getLogger(AktoMixpanel.class);
```
aktoboy, line 17: Minor: remove this object if not being used

aktoboy (approved):

PR #793, opened 2023-12-25
Feature/dependency flow

avneesh-akto:


apps/api-runtime/src/main/java/com/akto/dependency/DependencyAnalyser.java
```diff
@@ -0,0 +1,272 @@
...
+    Store urlParamValueStore; // this is to store all the url$param$value seen in response payload
+
+    Map<String, Set<String>> urlsToResponseParam = new HashMap<>();
+
+    Map<Integer, DependencyNode> nodes = new HashMap<>();
+    public Map<Integer, APICatalog> dbState = new HashMap<>();
+
+
+    public DependencyAnalyser(Map<Integer, APICatalog> dbState) {
+        valueStore = new HashSetStore(10_000);
```
ayushaga14, line 31: 10000 is too low for valuestore

apps/api-runtime/src/main/java/com/akto/dependency/DependencyAnalyser.java
```diff
@@ -0,0 +1,272 @@
...
+
+    public boolean urlValSeen(String url, Object val) {
+        return urlValueStore.contains(url + "$" + val.toString());
+    }
+
+    public boolean urlParamValueSeen(String url, String param, Object val) {
+        return urlParamValueStore.contains(url + "$" + param + "$" + val.toString());
+    }
+
+    public void add(String combinedUrlResp, String paramResp, String combinedUrlReq, String paramReq) {
```
ayushaga14, line 154: rename add

apps/api-runtime/src/main/java/com/akto/dependency/DependencyAnalyser.java
```diff
@@ -0,0 +1,272 @@
...
+
+                bulkUpdates.add(updateOneModel1);
+                bulkUpdates.add(updateOneModel2);
+                bulkUpdates.add(updateOneModel3);
+
+            }
+        }
+
+        // ordered has to be true or else won't work
+        if (bulkUpdates.size() > 0) DependencyNodeDao.instance.getMCollection().bulkWrite(bulkUpdates, new BulkWriteOptions().ordered(true));
```
ayushaga14, line 267: if there are n nodes, instead of ordered, can we insert findoneandupdate for 3 queries for a doc(which are ordered) in bulklist, and run unordered bulkupdate on this list of size n?

PR #795, opened 2024-01-01
bunch of postman fixes

ankush-jain-akto:
1. use url from originalRequest if present
2. ignore plus-sign for number comparison
3. add INTEGER/UUID/ALPHA_NUM first time itself
4. ignore status code for postman import
5. use sample-url for private-var-context


libs/dao/src/main/java/com/akto/dto/type/URLTemplate.java
```diff
@@ -75,6 +75,9 @@ public boolean match(String[] url, Method urlMethod) {
                         if (!"true".equals(thatToken.toLowerCase()) && !"false".equals(thatToken.toLowerCase())) return false;
                         break;
                     case INTEGER:
+                        if (thatToken.charAt(0) == '+') {
+                            thatToken = thatToken.substring(1);
+                        }
                         if (!NumberUtils.isParsable(thatToken) || thatToken.contains(".")) return false;
                         break;
                     case FLOAT:
```
aktoboy, line 83: Shouldn't we eat the plus sign for floats as well?

apps/api-runtime/src/main/java/com/akto/parsers/HttpCallParser.java
```diff
@@ -267,6 +267,10 @@ public List<HttpResponseParams> filterHttpResponseParams(List<HttpResponseParams
             }
 
             boolean cond = HttpResponseParams.validHttpResponseCode(httpResponseParam.getStatusCode());
+            if (httpResponseParam.getSource().equals(HttpResponseParams.Source.POSTMAN)) {
```
aktoboy, line 270: We will allow non 2xx responses to be ingested using postman?

aktoboy (approved):

apps/api-runtime/src/main/java/com/akto/parsers/HttpCallParser.java
ankush-jain-akto: good catch. Fixing this.

ankush-jain-akto:
https://github.com/akto-api-security/akto/pull/811 - merging as part of this PR

PR #797, opened 2024-01-06
Feature/default payloads

ankush-jain-akto:


libs/utils/src/main/java/com/akto/runtime/RuntimeUtil.java
```diff
@@ -0,0 +1,42 @@
...
+        try {
+            Map<String, List<String>> reqHeaders = httpResponseParams.getRequestParams().getHeaders();
+            List<String> host = reqHeaders.getOrDefault("host", new ArrayList<>());
+
+            String testHost = "";
+            if (host != null && !host.isEmpty() && host.get(0) != null) {
+                testHost = host.get(0);
+            } else {
+                String urlStr = httpResponseParams.getRequestParams().getURL();
+                URL url = new URL(urlStr);
```
avneesh-akto, line 25: use URI instead of URL

avneesh-akto (approved):

PR #812, opened 2024-01-12
Feature test chaining fixes

ayushaga14:


apps/testing/src/main/java/com/akto/test_editor/filter/FilterAction.java
```diff
@@ -1053,60 +1053,6 @@ public BasicDBObject getPrivateResourceCount(OriginalHttpRequest originalHttpReq
         BasicDBObject resp = new BasicDBObject();
         int privateCnt = 0;
         List<BasicDBObject> privateValues = new ArrayList<>();
-        if (APICatalog.isTemplateUrl(url)) {
```
ankush-jain-akto, line 1056: I know what this block did. Why we removed this @ayushaga14 ?

ankush-jain-akto (approved):

PR #824, opened 2024-01-16
Feature username enumeration

ayushaga14:


libs/dao/src/main/java/com/akto/dao/context/Context.java
```diff
@@ -76,6 +76,10 @@ public static int now() {
         return (int) (System.currentTimeMillis()/1000l);
     }
 
+    public static int nowInMillis() {
+        return (int) System.currentTimeMillis();
```
avneesh-akto, line 80: use long. this might exceed integer limits

apps/testing/src/main/java/com/akto/testing/workflow_node_executor/YamlNodeExecutor.java
```diff
@@ -128,6 +145,59 @@ public NodeResult processNode(Node node, Map<String, Object> varMap, Boolean all
...
+    }
+
+    public Map<String, Double> calcStats(List<Integer> arr) {
+        Map<String, Double> m = new HashMap<>();
+        if (arr.size() == 0) {
+            return m;
+        }
+        double last = 0.0;
+        double median;
+        for (int i = 0; i < arr.size(); i++) {
```
avneesh-akto, line 178: why loop? just do arr[arr.len - 1]

avneesh-akto (approved):

PR #838, opened 2024-01-24
add expired page

notshivansh:


apps/dashboard/src/main/java/com/akto/utils/billing/OrganizationUtils.java
```diff
@@ -276,34 +277,37 @@ public static BasicDBList fetchEntitlements(String orgId, String adminEmail) {
...
+        BasicDBObject reqBody = new BasicDBObject("orgId", orgIdUUID).append("adminEmail", adminEmail);
+        BasicDBObject orgMetaData = fetchFromBillingService("fetchOrgMetaData", reqBody);
+        orgMetaData = orgMetaData == null ? new BasicDBObject() : orgMetaData;
+        BasicDBObject additionalMetaData = (BasicDBObject) orgMetaData.getOrDefault("additionalMetaData", new BasicDBObject());
+        return additionalMetaData;
+    }
+
+    public static boolean fetchExpired(BasicDBObject additionalMetaData) {
+        String expiredStr = (String) additionalMetaData.getOrDefault(OrgMetaData.EXPIRED.name(), "false");
+        boolean expired = Boolean.parseBoolean(expiredStr);
```
aktoboy, line 310: Better check would be `expiredStr.equalsIgnoreCase('true')`.
Current code might break if some one sets an incorrect value for "EXPIRED" key.

apps/dashboard/web/polaris_web/web/src/apps/main/index.js
```diff
@@ -6,15 +6,38 @@ import { AppProvider } from "@shopify/polaris";
...
 
+let expired = false;
 
-root.render(
+if (window.DASHBOARD_MODE == 'ON_PREM' && (
+  !window.STIGG_CUSTOMER_ID ||
+  window.EXPIRED == 'true' ||
+  !window.STIGG_FEATURE_WISE_ALLOWED ||
+  (typeof window.STIGG_FEATURE_WISE_ALLOWED) !== 'object' ||
+  Object.keys(window.STIGG_FEATURE_WISE_ALLOWED).length == 0)) {
```
aktoboy, line 21: initially this obj can be empty, right?

aktoboy (changes requested):

apps/dashboard/web/src/apps/main/main.js
```diff
@@ -10,10 +12,29 @@ Vue.use(VuePlugin);
...
-  vuetify,
-  router,
-  store,
-  render: h => h(App)
-})
+let expired = false;
+
+if (window.DASHBOARD_MODE == 'ON_PREM' && (
+  !window.STIGG_CUSTOMER_ID ||
+  !window.EXPIRED ||
```
ankush-jain-akto, line 19: We shouldn't introduce this check int current release.

apps/dashboard/web/polaris_web/web/src/apps/main/index.js
```diff
@@ -6,15 +6,36 @@ import { AppProvider } from "@shopify/polaris";
...
 
 const container = document.getElementById("root");
 const root = createRoot(container);
 
+let expired = false;
 
-root.render(
+if (window.DASHBOARD_MODE == 'ON_PREM' && (
+  !window.STIGG_CUSTOMER_ID ||
+  !window.EXPIRED ||
```
ankush-jain-akto, line 18: We shouldn't do this in the current release

apps/dashboard/src/main/java/com/akto/filter/UserDetailsFilter.java
```diff
@@ -289,6 +287,29 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+
+            if (organization == null) {
+                httpServletResponse.sendError(403);
+                return;
+            }
+
+            if (organization.checkExpirationWithAktoSync()) {
+
+                // attempt to sync with billing once more
+                organization = InitializerListener.fetchAndSaveFeatureWiseAllowed(organization);
```
ankush-jain-akto, line 304: Is this a network call that will be hit every time an API is fired?

PR #847, opened 2024-02-05
Feature/dependency flow

avneesh-akto:


libs/utils/src/main/java/com/akto/util/HTTPHeadersExample.java
ankush-jain-akto, line 1: Rename this file to "StandardHTTPHeaders.java" or similar.

libs/dao/src/test/java/com/akto/utils/TestJsonUtils.java
```diff
@@ -1,92 +0,0 @@
...
-import org.junit.Test;
-
-import java.util.*;
-
-import static org.junit.Assert.*;
-
-public class TestJsonUtils {
-
-    @Test
-    public void testModify() {
```
ankush-jain-akto, line 18: Why delete the tests? Have we moved it somewhere?

PR #851, opened 2024-02-05
Feature/new onboarding

Ark2307:


apps/dashboard/src/main/java/com/akto/action/ApiCollectionsAction.java
```diff
@@ -373,6 +367,16 @@ public String fetchTimersInfo(){
         return Action.ERROR.toUpperCase();
     }
 
+    public String fetchCustomerEndpoints(){
+        try {
+            this.usageEndpoints = UsageMetricCalculator.calculateActiveEndpoints(0);
```
ankush-jain-akto, line 372: Hi @Ark2307 - we shouldn't recalculate here - we should simply fetch last usage data from database

apps/dashboard/src/main/java/com/akto/action/testing/StartTestAction.java
```diff
@@ -713,6 +712,23 @@ public String deleteTestDataFromSummaryId(){
...
+    private int testRunsByUser;
+
+    private int getUserTestingRuns(){
+        Bson filter = Filters.ne(TestingRun.NAME, "Onboarding demo test");
+        return (int) TestingRunDao.instance.getMCollection().countDocuments(filter);
+    }
+
+    public String getUsageTestRuns(){
+        try {
+            this.testRunsByUser = getUserTestingRuns();
```
ankush-jain-akto, line 724: You need to fetch just yes-or-no - not the entire count of testing runs.

libs/dao/src/main/java/com/akto/dao/SingleTypeInfoDao.java
```diff
@@ -77,6 +77,13 @@ public void createIndicesIfAbsent() {
         fieldNames =  new String[]{SingleTypeInfo._RESPONSE_CODE, SingleTypeInfo.SUB_TYPE, SingleTypeInfo._TIMESTAMP};
         MCollection.createIndexIfAbsent(getDBName(), getCollName(), fieldNames, true);
 
+        // needed for usage metric calculation
+        MCollection.createIndexIfAbsent(getDBName(), getCollName(),
```
ankush-jain-akto, line 81: Please remove indices. We should find a simpler way to calculate metric.

PR #873, opened 2024-02-11
diffs

ankush-jain-akto:


apps/api-analyser/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM openjdk
```
ayushaga14, line 1: should we remove api analyser?

ankush-jain-akto:
Already merged in a diff PR

PR #879, opened 2024-02-14
handle long in url param

avneesh-akto:


libs/dao/src/main/java/com/akto/dto/type/SingleTypeInfo.java
```diff
@@ -143,6 +144,11 @@ public enum SuperType {
         BOOLEAN, INTEGER, FLOAT, STRING, NULL, OTHER, CUSTOM, OBJECT_ID
     }
 
+    public static void main(String[] args) {
```
notshivansh, line 147: is this needed ?

PR #903, opened 2024-02-22
Upload functionality in editor

Ark2307:


apps/testing/src/main/java/com/akto/test_editor/execution/Executor.java
```diff
@@ -420,8 +420,11 @@ private ExecutorSingleOperationResp modifyAuthTokenInRawApi(TestRoles testRole,
         return null;
     }
 
+
     public ExecutorSingleOperationResp runOperation(String operationType, RawApi rawApi, Object key, Object value, Map<String, Object> varMap, AuthMechanism authMechanism, List<CustomAuthType> customAuthTypes) {
         switch (operationType.toLowerCase()) {
+            case "attach_file":
+                return Operations.addHeader(rawApi, "x-attach_file" , key.toString());
```
ayushaga14, line 427: x-akto-attach-file

apps/testing/src/main/java/com/akto/testing/ApiExecutor.java
```diff
@@ -201,6 +205,44 @@ private static OriginalHttpResponse getRequest(OriginalHttpRequest request, Requ
...
+                    @Override
+                    public void writeTo(BufferedSink sink) throws IOException {
+                        byte[] chunk = new byte[CHUNK_SIZE];
+                        int bytesRead;
+                        long totalBytesRead = 0; 
+                        long maxBytes = 100L * 1024 * 1024; 
+                        while ((bytesRead = urlInputStream.read(chunk)) != -1) {
+                            totalBytesRead += bytesRead;
+                            if (totalBytesRead > maxBytes) {
+                                throw new IOException("File size exceeds the maximum limit of 100MB");
```
ayushaga14, line 231: instead of throwing exception, let's exit from the function, so that we just upload 100Mb of data at max

apps/testing/src/main/java/com/akto/testing/HTTPClientHandler.java
```diff
@@ -55,11 +55,10 @@ static class NormalResponseInterceptor implements Interceptor {
         @Override
         public @NotNull Response intercept(Chain chain) throws IOException {
             Request request = chain.request();
-
             try {
                 Buffer buffer = new Buffer();
                 RequestBody requestBody = request.body();
-                if (requestBody != null) {
+                if (requestBody != null && requestBody.contentLength() != -1) {
```
ayushaga14, line 61: can any other kind of request apart from file upload have requestBody.contentLength() != -1

apps/testing/src/main/java/com/akto/testing/ApiExecutor.java
```diff
@@ -201,6 +205,44 @@ private static OriginalHttpResponse getRequest(OriginalHttpRequest request, Requ
...
+            InputStream urlInputStream = sourceFileUrl.openStream();
+            final int CHUNK_SIZE = 1 * 1024 * 1024 ;
+
+            RequestBody requestBody = new MultipartBody.Builder()
+                .setType(MultipartBody.FORM)
+                .addFormDataPart("file", "filename", new RequestBody() {
+                    @Override
+                    public MediaType contentType() {
+                        return Utils.getMediaType(fileUrl);
+                    }
```
ayushaga14, line 220: is this being used?

ayushaga14 (changes requested):
check comments

apps/testing/src/main/java/com/akto/test_editor/execution/Executor.java
Ark2307: done

apps/testing/src/main/java/com/akto/testing/ApiExecutor.java
Ark2307: done

apps/testing/src/main/java/com/akto/testing/HTTPClientHandler.java
Ark2307: not in our case, because we have the body of type string which has fixed length

ayushaga14 (approved):

PR #964, opened 2024-03-30
changes from mono to akto

avneesh-akto:


ankush-jain-akto (commented):
The behaviours around sending invitationEmail etc. were different between mono and akto. Check if diff is maintained

ayushaga14 (approved):

PR #979, opened 2024-04-05
add indexing in users, accounts and singletypeinfo docs

Addy-codes:
Added 3 new indices in the following documents:
1. common.users
2. common.accounts
3. 1000000.single_type_info


ankush-jain-akto (approved):

libs/dao/src/main/java/com/akto/dao/SingleTypeInfoDao.java
```diff
@@ -52,8 +52,8 @@ public void createIndicesIfAbsent() {
         String[] fieldNames = {SingleTypeInfo._URL, SingleTypeInfo._METHOD, SingleTypeInfo._RESPONSE_CODE, SingleTypeInfo._IS_HEADER, SingleTypeInfo._PARAM, SingleTypeInfo.SUB_TYPE, SingleTypeInfo._API_COLLECTION_ID};
```
ayushaga14, line 52: are we not removing the response code index we discussed

Addy-codes: I was told by Ankush not to remove any existing indices for now.

ayushaga14 (approved):

PR #1060, opened 2024-04-27
feat: send slack alerts on user invitations

TangoBeee:


libs/utils/src/main/java/com/akto/notifications/slack/SlackSender.java
```diff
@@ -0,0 +1,73 @@
...
+import com.slack.api.Slack;
+import com.slack.api.webhook.WebhookResponse;
+
+import java.util.List;
+
+public class SlackSender {
+    private static final LoggerMaker loggerMaker = new LoggerMaker(SlackAlerts.class, LoggerMaker.LogDb.DASHBOARD);
+
+    public static void sendAlert(int accountId, SlackAlerts alert) {
+        Thread thread = new Thread(() -> {
```
avneesh-akto, line 17: I realised we should use `Executors.newFixedThreadPool`, creating raw threads is expensive in java world

libs/utils/src/main/java/com/akto/notifications/slack/SlackSender.java
```diff
@@ -0,0 +1,73 @@
...
+            final int[] retryDelays = {1000, 5000, 25000};
+
+            while(attempts < retryDelays.length + 1) {
+                try {
+                    WebhookResponse response = slack.send(webhookUrl, payload);
+                    int statusCode = response.getCode();
+
+                    if(statusCode == 200) {
+                        loggerMaker.infoAndAddToDb("Slack Alert Type: " + alertType + " Info: " + "Alert sent successfully.");
+                        Thread.currentThread().interrupt();
```
avneesh-akto, line 45: Just use `return` no need to interrupt the thread, interrupting the thread makes sense if you are waiting for a function to complete like thread.sleep() etc.

PR #1065, opened 2024-04-29
Feature/graphql test improvements

shivam-rawat-akto:


apps/testing/src/main/java/com/akto/test_editor/Utils.java
```diff
@@ -320,6 +320,81 @@ public static BasicDBObject fetchJsonObjForString(Object val) {
...
+            node = mapper.readTree(jp);
+        } catch (IOException e) {
+            return values;
+        }
+
+        findAllKeys(node, key, values, false);
+        return values;
+    }
+
+    public static void main(String[] args) {
```
ayushaga14, line 347: remove if not required

apps/testing/src/main/java/com/akto/test_editor/execution/VariableResolver.java
```diff
@@ -665,7 +665,7 @@ public static List<String> fetchWordList(Map<ApiInfoKey, List<String>> modifiedS
         }
         for (String s : wordListSet) {
             wordListVal.add(s);
-            if (wordListVal.size() >= 10) {
+            if (wordListVal.size() >= 10 && !"terminal_keys".equals(location)) {
```
ayushaga14, line 668: should we put some limit on graphql wordlist separately? this can get huge

apps/testing/src/main/java/com/akto/test_editor/filter/FilterAction.java
```diff
@@ -285,6 +285,9 @@ public DataOperandsFilterResponse applyFilterOnPayload(FilterActionRequest filte
 
         BasicDBObject payloadObj = new BasicDBObject();
         try {
+            if (payload != null && payload.startsWith("[")) {
```
ayushaga14, line 288: i think we would need this in extract method as well for req/resp payloads, can you check once

libs/utils/src/main/java/com/akto/graphql/GraphQLUtils.java
```diff
@@ -167,42 +238,54 @@ public TraversalControl visitField(Field node, TraverserContext<Node> context) {
...
-        return gson.toJson(payloadList);
+        if (payloadObj instanceof Object[]) {
+            return gson.toJson(payloadList);
+        } else {
+            return gson.toJson(payloadList[0]);
+        }
+    }
+
+
+    public static void main(String[] args) {
```
ayushaga14, line 254: remove

ayushaga14 (changes requested):
check comments

ayushaga14 (approved):

PR #1083, opened 2024-05-04
feat: created an api test status alert

TangoBeee:


apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -485,6 +472,37 @@ private static void raiseMixpanelEvent(ObjectId summaryId, TestingRun testingRun
             }
             int val = severityCount.get(key);
             severityCount.put(key, val+1);
+            apisAffectedCount.put(testingRunIssues.getId().getTestSubCategory(), apisAffectedCount.getOrDefault(testingRunIssues.getId().getTestSubCategory(), 0)+1);
```
avneesh-akto, line 475: code cleanup plz. make it a variable and then reuse it.

apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -495,6 +513,39 @@ private static void raiseMixpanelEvent(ObjectId summaryId, TestingRun testingRun
...
+        if(testingEndpoints.getType().equals(TestingEndpoints.Type.COLLECTION_WISE)) {
+            CollectionWiseTestingEndpoints collectionWiseTestingEndpoints = (CollectionWiseTestingEndpoints) testingEndpoints;
+            int apiCollectionId = collectionWiseTestingEndpoints.getApiCollectionId();
+            ApiCollection apiCollection = ApiCollectionsDao.instance.getMeta(apiCollectionId);
+            collection = apiCollection.getName();
+        }
+
+        long currentTime = Context.now();
+        long startTimestamp = testingRunResultSummary.getStartTimestamp();
+        long scanTimeInSeconds = Math.abs(currentTime - startTimestamp);
```
avneesh-akto, line 529: use testEndtime instead of currentTime

PR #1091, opened 2024-05-07
made changes to calculate risk score correctly in optimised way

Ark2307:


apps/dashboard/src/main/java/com/akto/utils/RiskScoreOfCollections.java
```diff
@@ -246,4 +247,37 @@ public void mapSensitiveSTIsInApiInfo(int timeStampFilter,int cronTime){
...
     }
 
+    public void calculateRiskScoreForAllApis() {
+        int timeStamp = Context.now() - 24*60*60;
+        int limit = 1000;
+        int count = 0; 
+
+        List<WriteModel<ApiInfo>> bulkUpdates = new ArrayList<>();
+        Bson filter = Filters.or(
+            Filters.not(Filters.exists(ApiInfo.LAST_CALCULATED_TIME)),
```
avneesh-akto, line 257: Check if this is using index. If not simplify using 
filters.and(filters.exists(false), Filters.gt() )

libs/dao/src/main/java/com/akto/dao/ApiInfoDao.java
```diff
@@ -71,7 +74,7 @@ public void createIndicesIfAbsent() {
                 new String[] {ApiInfo.RISK_SCORE }, false);
 
         MCollection.createIndexIfAbsent(getDBName(), getCollName(),
-                new String[] {ApiInfo.ID_API_COLLECTION_ID, ApiInfo.RISK_SCORE }, false);
+                new String[] { ApiInfo.RISK_SCORE, ApiInfo.ID_API_COLLECTION_ID }, false);
```
avneesh-akto, line 77: why was the order reversed

avneesh-akto (approved):

PR #1104, opened 2024-05-14
fix: create 3 different modes for Dashboard, TestEditor and RunTest, …

Addy-codes:
…remove YAML-content where it's not needed


apps/dashboard/src/main/java/com/akto/action/testing_issues/IssuesAction.java
```diff
@@ -225,10 +225,29 @@ public static BasicDBObject createSubcategoriesInfoObj(TestConfig testConfig) {
...
+                vulnerableRequests = VulnerableRequestForTemplateDao.instance.findAll(Filters.empty());
+                break;
+            case "Dashboard":
+                testSourceConfigs = TestSourceConfigsDao.instance.findAll(Filters.empty());
+                break;
+            default:
+                return "ERROR: Invalid mode";
+        }
+
+        Map<String, TestConfig> testConfigMap = YamlTemplateDao.instance.fetchTestConfigMap(includeYamlContent,
```
Ark2307, line 249: See YamlTemplateDao.java, there we parse the template before using includeYamlContent's value, this means we will just not send it to the UI but it will still parse it, thus modify that file also

apps/dashboard/src/main/java/com/akto/action/testing_issues/IssuesAction.java
```diff
@@ -225,10 +225,29 @@ public static BasicDBObject createSubcategoriesInfoObj(TestConfig testConfig) {
...
+            case "RunTests":
+                categories = GlobalEnums.TestCategory.values();
+                break;
+            case "TestEditor":
+                includeYamlContent = true;
+                vulnerableRequests = VulnerableRequestForTemplateDao.instance.findAll(Filters.empty());
+                break;
+            case "Dashboard":
+                testSourceConfigs = TestSourceConfigsDao.instance.findAll(Filters.empty());
+                break;
```
Ark2307, line 244: default case should be like it was working before

PR #1137, opened 2024-05-23
Feature/improve testing UI

Ark2307:


apps/testing/src/main/java/com/akto/testing/TestExecutor.java
```diff
@@ -578,31 +547,36 @@ public void startTestNew(ApiInfo.ApiInfoKey apiInfoKey, ObjectId testRunId,
...
-            } catch (Exception e) {
-                loggerMaker.errorAndAddToDb("Error while running tests for " + testSubCategory +  ": " + e.getMessage(), LogDb.TESTING);
-                e.printStackTrace();
-            }
-            if (testingRunResult != null) {
-                testingRunResults.add(testingRunResult);
-                countSuccessfulTests++;
+                insertResultsAndMakeIssues(testingRunResults, testRunResultSummaryId);
+            }else{
+                logger.info("Test stopped for id: " + testRunId.toString());
```
ayushaga14, line 577: add to loggermaker and add accountid

apps/testing/src/main/java/com/akto/testing_issues/TestingIssuesHandler.java
```diff
@@ -101,15 +107,34 @@ private void insertVulnerableTestsIntoIssuesCollection(List<WriteModel<TestingRu
                                                            Map<TestingIssuesId, TestingRunResult> testingIssuesIdsMap,
                                                            List<TestingRunIssues> testingRunIssuesList) {
         int lastSeen = Context.now();
+        AtomicReference<ObjectId> summaryIdRef = new AtomicReference<>(null);
```
avneesh-akto, line 110: please use simple for loop

PR #1200, opened 2024-06-17
Add: header and footer images which static texts

Addy-codes:


oren-akto:
Remove header and footer text from jsx file and css

PR #1324, opened 2024-08-02
make merge fn async

notshivansh:


apps/mini-runtime/src/main/java/com/akto/hybrid_parsers/HttpCallParser.java
```diff
@@ -166,7 +170,24 @@ public void syncFunction(List<HttpResponseParams> responseParams, boolean syncIm
...
             this.last_synced = Context.now();
             this.sync_count = 0;
-            MergeLogicLocal.mergingJob(apiCatalogSync.dbState);
+            /*
+             * submit a job only if it is not running.
+             */
+            if (!pgMerging) {
+                int accountId = Context.accountId.get();
+                service.submit(() -> {
+                    pgMerging = true;
```
aktoboy, line 179: Move this line out of lambda

notshivansh: fixed

apps/mini-runtime/src/main/java/com/akto/hybrid_parsers/HttpCallParser.java
```diff
@@ -166,7 +170,24 @@ public void syncFunction(List<HttpResponseParams> responseParams, boolean syncIm
...
             this.sync_count = 0;
-            MergeLogicLocal.mergingJob(apiCatalogSync.dbState);
+            /*
+             * submit a job only if it is not running.
+             */
+            if (!pgMerging) {
+                int accountId = Context.accountId.get();
+                pgMerging = true;
+                service.submit(() -> {
+                    Context.accountId.set(accountId);
```
aktoboy, line 180: Write entire thread inside a try-catch and set pgMerging to false inside a finally block

PR #1348, opened 2024-08-13
Feature/usage cyborg

notshivansh:


libs/utils/src/main/java/com/akto/data_actor/DbLayer.java
```diff
@@ -632,6 +635,16 @@ public static TestingRunResultSummary updateIssueCountInSummary(String summaryId
                 options);
     }
 
+    public static List<Integer> fetchDeactivatedCollections() {
```
ayushaga14, line 638: fix or add a todo for adding projection in mongo query

PR #1356, opened 2024-08-14
calc access type when updated

notshivansh:


apps/dashboard/src/main/java/com/akto/action/AdminSettingsAction.java
```diff
@@ -243,6 +243,23 @@ public String editPrivateCidrList(){
             AccountSettingsDao.instance.getMCollection().updateOne(
                 AccountSettingsDao.generateFilter(), Updates.set(AccountSettings.PRIVATE_CIDR_LIST, privateCidrList)
             );
+
+            int accountId = Context.accountId.get();
```
notshivansh, line 247: This job will run everytime the user presses enter. Fix: add an apply button in UI instead of making an API call everytime.

PR #1372, opened 2024-08-19
domain unreachable in test run results

shivam-rawat-akto:


apps/dashboard/src/main/java/com/akto/action/testing/StartTestAction.java
```diff
@@ -531,17 +530,23 @@ public String fetchTestingRunResults() {
                 case VULNERABLE:
                     testingRunResultFilters.add(Filters.eq(TestingRunResult.VULNERABLE, true));
```
Ark2307, line 531: Add filters.nin in vulnerable otherwise it will show SKIPPED_EXEC_API_REQUEST_FAILED these also

apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/observe/api_collections/RunTest.jsx
```diff
@@ -270,7 +270,7 @@ function RunTest({ endpoints, filtered, apiCollectionId, disabled, runTestFromOu
         return abc
     }, [])
 
-    const testRunTimeOptions = [{ label: "Till complete", value: "Till complete" }, ...runTimeMinutes, ...runTimeHours]
```
Ark2307, line 273: what is this?

Ark2307 (changes requested):

apps/dashboard/src/main/java/com/akto/action/testing/StartTestAction.java
shivam-rawat-akto: check line 534 535, 
did that exactly

shivam-rawat-akto: also, 
for this case, no test will be vulnerable, as api request failed,

Ark2307 (approved):

PR #1452, opened 2024-09-06
Feature/source code analyser

shivam-rawat-akto:


pom.xml
```diff
@@ -23,6 +23,7 @@
     <modules>
         <module>apps</module>
         <module>libs</module>
+        <module>apps/source-code-analyser</module>
```
ankush-jain-akto, line 26: You don't need this @shivam-rawat-akto . 

You should add `source-code-analyser` in pom.xml of `apps` folder

ankush-jain-akto (approved):

PR #1491, opened 2024-09-13
Sensitive key redaction

ayushaga14:


apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/APICatalogSync.java
```diff
@@ -1601,7 +1601,19 @@ public DbUpdateReturnHybrid getDBUpdatesForParamsHybrid(APICatalog currentDelta,
...
             updates.add(updatePayload.toString());
 
-            if (!redactSampleData && deltaInfo.getExamples() != null && !deltaInfo.getExamples().isEmpty()) {
+            if (!(redactSampleData || collectionLevelRedact) && deltaInfo.getExamples() != null && !deltaInfo.getExamples().isEmpty()) {
+                Set<Object> updatedSampleData = new HashSet<>();
+                for (Object example : deltaInfo.getExamples()) {
+                    try{
+                        String exampleStr = (String) example;
+                        String s = RedactSampleData.redactDataTypes(exampleStr);
+                        loggerMaker.infoAndAddToDb("redacted data, updated sample is " + s, LogDb.RUNTIME);
```
notshivansh, line 1610: inside this function, later on context.accountId.get is being used. we need to make sure that the correct accountId is set or is atleast is being used for that block.

PR #1492, opened 2024-09-13
Handle redacted field

ayushaga14:


notshivansh (approved):

libs/dao/src/main/java/com/akto/dto/CustomDataTypeMapper.java
```diff
@@ -20,6 +20,7 @@ public class CustomDataTypeMapper {
     Conditions valueConditions;
     Conditions.Operator operator;
     private IgnoreData ignoreData;
+    private Boolean redacted;
```
notshivansh, line 23: we need to check if adding a field does not break the existing mini-runtimes without this field.

PR #1495, opened 2024-09-16
Added db calls for mini-runtime

Ark2307:


libs/utils/src/main/java/com/akto/data_actor/DataActor.java
```diff
@@ -235,4 +235,6 @@ public abstract class DataActor {
     public abstract List<YamlTemplate> fetchFilterYamlTemplates();
 
     public abstract void insertProtectionLog(Log log);
+    
+    public abstract List<YamlTemplate> fetchActiveAdvancedFilters();
```
notshivansh, line 239: Since this is an abstract class, all implementations of the abstract class must implement all the functions of the same. **THIS CODE WILL NOT COMPILE.** Please run the code once, before submitting.

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -1966,6 +1967,15 @@ public String fetchFilterYamlTemplates() {
         return Action.SUCCESS.toUpperCase();
     }
 
+    public String fetchActiveAdvancedFilters(){
+        try {
+            this.activeAdvancedFilters = DbLayer.fetchActiveFilterTemplates();
+        } catch (Exception e) {
+            return Action.ERROR.toUpperCase();
```
notshivansh, line 1974: add error logs for monitoring.

libs/utils/src/main/java/com/akto/data_actor/DataActor.java
Ark2307: It's compiling, refresh once
and dont take assumption please before commenting

notshivansh: Screenshot 2024-09-16 at 9 37 40 AM

well, it was not when I reviewed, not assuming here.

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
ayushaga14: resolve comments ^

PR #1583, opened 2024-10-04
Feature/datatype to test

notshivansh:


apps/dashboard/src/main/java/com/akto/action/test_editor/SaveTestEditorAction.java
```diff
@@ -289,7 +290,9 @@ public String runTestForGivenTemplate() {
         }
 
         try {
-            GlobalEnums.Severity.valueOf(testConfig.getInfo().getSeverity());
+            if (!TestConfig.DYNAMIC_SEVERITY.equals(testConfig.getInfo().getSeverity())) {
+                GlobalEnums.Severity.valueOf(testConfig.getInfo().getSeverity());
```
Ark2307, line 294: can't we add another value DYNAMIC_SEVERITY in globalenums under severity?

apps/dashboard/src/main/java/com/akto/action/LoginAction.java
```diff
@@ -190,6 +191,22 @@ public static String loginUser(User user, HttpServletResponse servletResponse, b
...
                         )
                 );
+                /*
+                 * Creating datatype to template on user login.
+                 * TODO: Remove this job once templates for majority users are created.
+                 */
+                service.submit(() -> {
+                    try {
+                        for (String accountIdStr : user.getAccounts().keySet()) {
+                            int accountId = Integer.parseInt(accountIdStr);
```
Ark2307, line 201: add try catch here and then handle exception downwards. sometimes this accountIdStr gives null

apps/dashboard/src/main/java/com/akto/action/test_editor/SaveTestEditorAction.java
notshivansh: Nope, since the values for severity there are shown in the UI and used actively in documents being stored in the db,  whearas dynamic_severity is only present at the test template level and not used anywhere actively, thus makes more sense to make it a seprate entity.

apps/dashboard/src/main/java/com/akto/action/LoginAction.java
notshivansh: There is a try catch for the entire job. accountIdStr is coming from the keyset. In case the keyset itself is null, gets handled here. In case the keyset contains something like this [ 1234, 2345, null, 3456 ] , in this case the logic would break. should we add a try catch inside the for loop as well, didn't seem necessary since there was one outside ?

libs/dao/src/main/java/com/akto/dao/testing/TestingRunResultDao.java
```diff
@@ -63,35 +68,69 @@ public List<TestingRunResult> fetchLatestTestingRunResult(Bson filters, int limi
...
-                .projection(projections)
-                .sort(Sorts.descending("_id"))
-                .skip(skip)
-                .limit(limit)
-                .cursor();
+        List<Bson> pipeline = new ArrayList<>();
+        pipeline.add(Aggregates.match(filters));
+        pipeline.add(Aggregates.project(projections));
+        pipeline.add(Aggregates.sort(Sorts.descending(Constants.ID)));
+        pipeline.add(Aggregates.skip(skip));
```
Ark2307, line 85: projections or aggregate doesnt have unique key, use TestingRunResult instead of BasicDBObject

Ark2307 (approved):

PR #1642, opened 2024-10-17
delete non vulnerable test results near cap

notshivansh:


apps/testing/src/main/java/com/akto/testing/Main.java
```diff
@@ -453,6 +457,34 @@ public void run() {
...
+                 * since we cap on size as well as number of documents.
+                 */
+                deleteScheduler.execute(() -> {
+                    Context.accountId.set(accountId);
+                    try {
+                        /*
+                         * Since this is a heavy job,
+                         * running it for only one account on one instance at a time.
+                         */
+                        boolean dibs = Cluster.callDibs(Cluster.DELETE_TESTING_RUN_RESULTS, 60 * 60, 60);
```
avneesh-akto, line 473: if dibs is not acquired, then we will never trim the collection. Need to handle that.

avneesh-akto: acquire dibs post knowing that we have reached a limit

avneesh-akto (approved):

PR #1668, opened 2024-10-24
skip tests on host unreachable

notshivansh:


libs/utils/src/main/java/com/akto/testing/HostValidator.java
```diff
@@ -0,0 +1,113 @@
...
+            return false;
+        } finally {
+            if (response != null) {
+                response.close();
+            }
+        }
+        return true;
+    }
+
+    private static String getUniformUrl(String url){
```
avneesh-akto, line 73: write some unit tests for this:
1. with scheme, host and port
2. with scheme, host
3. only scheme

avneesh-akto (approved):

PR #1675, opened 2024-10-29
Feature/count new params

Ark2307:


apps/dashboard/src/main/java/com/akto/action/observe/InventoryAction.java
```diff
@@ -712,6 +712,50 @@ public String fetchChanges() {
 
         return Action.SUCCESS.toUpperCase();
     }
+    public String fetchRecentParams (){
+
+        // ignore time for new params detection {15 days}
+        Bson apiInfoFilter = Filters.lt(ApiInfo.DISCOVERED_TIMESTAMP, this.startTimestamp - (Utils.DELTA_PERIOD_VALUE/4));
+
+        long countApiInfosValid = ApiInfoDao.instance.count(apiInfoFilter);
+        long countApiInfosInvalid = ApiInfoDao.instance.count(Filters.not(apiInfoFilter));
```
avneesh-akto, line 721: instead of using not
1. use Filters.gt (i.e having 2 different filters)
2. use estimatedCount and subtract from it

Not is generally not efficient

avneesh-akto (approved):

PR #1750, opened 2024-11-21
check for template updates every 24 hrs

notshivansh:


apps/dashboard/src/main/java/com/akto/action/LoginAction.java
```diff
@@ -190,29 +194,32 @@ public static String loginUser(User user, HttpServletResponse servletResponse, b
...
+                        } catch (Exception e) {
                         }
-                    } catch (Exception e) {
-                    }
-                });
+                    });
+                }
             }
             service.submit(() ->{
                 triggerVulnColUpdation(user);
```
Ark2307, line 225: can we do same here please

Ark2307 (changes requested):

apps/dashboard/src/main/java/com/akto/action/LoginAction.java
notshivansh: done

PR #1816, opened 2024-12-12
feat: now users can only delete their own test roles unless it's admin

TangoBeeAkto:


apps/dashboard/src/main/java/com/akto/action/testing/TestRolesAction.java
```diff
@@ -136,6 +139,18 @@ public String deleteTestRole() {
             return ERROR.toUpperCase();
         }
 
+        User user = getSUser();
+        if(user == null) {
+            addActionError("User not found.");
+            return ERROR.toUpperCase();
+        }
+
+        RBAC.Role currentRoleForUser = RBACDao.getCurrentRoleForUser(user.getId(), Context.accountId.get());
```
Ark2307, line 148: first check for user and then call role for user

Ark2307 (approved):

PR #1887, opened 2025-01-02
Fix/fix summaries after ignored test

Ark2307:


apps/dashboard/src/main/java/com/akto/action/testing_issues/IssuesAction.java
```diff
@@ -530,6 +539,42 @@ public String bulkUpdateIssueStatus () {
...
         }
         TestingRunIssuesDao.instance.updateMany(Filters.in(ID, issueIdArray), update);
+
+        int accountId = Context.accountId.get();
+        executorService.schedule( new Runnable() {
+            public void run() {
+                Context.accountId.set(accountId);
+                try {
+
+                    Map<String, Integer> countIssuesMap = new HashMap<>();
```
notshivansh, line 549: since this is being used as a starter, not meant to be modified, make it final/const.

apps/dashboard/src/main/java/com/akto/action/testing/StartTestAction.java
```diff
@@ -507,21 +509,25 @@ public String fetchTestingRunResultSummary() {
         }
     }
 
+    private static Bson vulnerableFilter = Filters.and(
+        Filters.eq(TestingRunResult.VULNERABLE, true),
+        Filters.in(TestingRunResult.IS_IGNORED_RESULT, Arrays.asList(null, false))
```
notshivansh, line 514: Filters.in(TestingRunResult.IS_IGNORED_RESULT, Arrays.asList(null, false)) instead of this use OR[ !exists , false ] query.

notshivansh (approved):

PR #1898, opened 2025-01-02
Fixed check for all accounts

Ark2307:


apps/dashboard/src/main/java/com/akto/utils/sso/SsoUtils.java
```diff
@@ -32,9 +35,21 @@ public static boolean isAnySsoActive(int accountId){
...
-        return ConfigsDao.instance.count(filter) > 0;
+        if(DashboardMode.isMetered() && !DashboardMode.isOnPremDeployment()){
+            int accountId = Context.accountId.get();
+            if(!isAnySsoActive(accountId)){
+                return ConfigsDao.instance.count(Filters.and(
+                    Filters.eq(Constants.ID, "OKTA-ankush"),
+                    Filters.eq(OktaConfig.ACCOUNT_ID, accountId)
+                )) > 0;
+            }else{
+                return true;
```
notshivansh, line 46: return false ?

notshivansh (approved):

PR #1911, opened 2025-01-03
moving okta-sso for new id

Ark2307:


libs/dao/src/main/java/com/akto/dto/Config.java
```diff
@@ -367,7 +367,7 @@ public static class OktaConfig extends Config {
 
         public OktaConfig() {
             this.configType = ConfigType.OKTA;
-            this.id = CONFIG_ID;
+            this.id = CONFIG_ID + "_" + this.accountId;
```
TangoBeeAkto, line 370: This won't work because the constructor will be called and the id variable will be initialized before the setAccountId method is called so the value will default to 0.

notshivansh (approved):

PR #2050, opened 2025-02-03
checking of overide url at frontend and backend both

amitpanwar789:


apps/dashboard/src/main/java/com/akto/action/testing/StartTestAction.java
```diff
@@ -1137,7 +1137,7 @@ public String modifyTestingRunConfig(){
             return Action.ERROR.toUpperCase();
         }
         try {
-            if (this.testingRunConfigId == 0) {
+            if (this.testingRunConfigId < 0) {
```
amitpanwar789, line 1140: testingRunConfigId can't be less than 0

PR #2132, opened 2025-02-23
chore:enable hot module replacement

devendershekhawat:
Enables HMR on frontend for faster development and good DX. Separate webpack config that does not mess with existing build process.

### DIFF
--- a/apps/dashboard/web/polaris_web/package.json
+++ b/apps/dashboard/web/polaris_web/package.json
@@ -43,7 +43,8 @@
     "zustand": "^4.3.8"
   },
   "scripts": {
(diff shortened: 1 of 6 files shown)


apps/dashboard/web/polaris_web/package.json
```diff
@@ -82,6 +83,7 @@
     "css-loader": "^6.8.1",
     "file-loader": "^6.2.0",
     "monaco-editor-webpack-plugin": "^7.1.0",
+    "react-refresh": "^0.16.0",
```
devendershekhawat, line 86: Help with HMR in deeply nested components

PR #2152, opened 2025-02-28
Test Roles Main Screen

notlazykid:


amitpanwar789:
it has a glitch, when we delete the role from action list, it will successfully delete the role but also run onRowClick function and move to test Role Detail page

Ark2307 (approved):

PR #2360, opened 2025-04-14
remove some new object init calls

ayushaga14:


apps/threat-detection/src/main/java/com/akto/threat/detection/tasks/MaliciousTrafficDetectorTask.java
```diff
@@ -147,16 +148,7 @@ private Map<String, FilterConfig> getFilters() {
...
-      VariableResolver.resolveWordList(
-          varMap,
-          new HashMap<ApiInfo.ApiInfoKey, List<String>>() {
-            {
-              put(apiInfoKey, Collections.singletonList(message));
-            }
-          },
-          apiInfoKey);
-
+      varMap.clear();
```
gauravakto, line 151: No need to do resolve var map ?

libs/utils/src/main/java/com/akto/test_editor/filter/FilterAction.java
```diff
@@ -48,6 +48,8 @@
 import static com.akto.runtime.utils.Utils.parseKafkaMessage;
 
 public final class FilterAction {
+
+    private final DataOperandFilterRequest dataOperandFilterRequest = new DataOperandFilterRequest();
```
gauravakto, line 52: this should be private static final??

libs/dao/src/main/java/com/akto/dto/test_editor/DataOperandFilterRequest.java
```diff
@@ -40,4 +40,11 @@ public void setOperand(String operand) {
         this.operand = operand;
     }
 
+    public DataOperandFilterRequest modify(Object data, Object queryset, String operand) {
```
gauravakto, line 43: This whole class is a subset of filterActionRequest with one extra field `data`.

What if we delete this whole class and just use the filterActionRequest ??

gauravakto (approved):
check minor comments, looks fine to me

PR #2364, opened 2025-04-15
Added throttle

Ark2307:


apps/mini-testing/src/main/java/com/akto/testing/kafka_utils/Producer.java
```diff
@@ -28,13 +28,24 @@
...
-    public static Void pushMessagesToKafka(List<SingleTestPayload> messages, AtomicInteger totalRecords){
+    
+    private static final LoggerMaker loggerMaker = new LoggerMaker(Producer.class, LogDb.TESTING);
+
+    public static Void pushMessagesToKafka(List<SingleTestPayload> messages, AtomicInteger totalRecords, AtomicInteger throttleNumber){
         for(SingleTestPayload singleTestPayload: messages){
             String messageString = singleTestPayload.toString();
-            producer.send(messageString, Constants.TEST_RESULTS_TOPIC_NAME, totalRecords);
+            try {
+                while (throttleNumber.get() > 500) {
```
ayushaga14, line 40: need to do this after we have sent for ex - 500 messages

ayushaga14 (approved):

PR #2394, opened 2025-04-21
feat: add support for multiple user environment types and update rela…

TangoBeeAkto:
…ted functionality

### DIFF
--- a/libs/dao/src/main/java/com/akto/dto/ApiCollection.java
+++ b/libs/dao/src/main/java/com/akto/dto/ApiCollection.java
@@ -14,8 +14,6 @@ import com.akto.dto.testing.CustomTestingEndpoints;
 import com.akto.dto.testing.TestingEndpoints;
 import com.akto.dto.traffic.CollectionTags;
 
-import lombok.Getter;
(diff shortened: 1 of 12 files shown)


libs/dao/src/main/java/com/akto/dto/ApiCollection.java
```diff
@@ -300,12 +313,25 @@ public void setSampleCollectionsDropped(boolean sampleCollectionsDropped) {
...
 
+    @Deprecated
 	public void setUserSetEnvType(String userSetEnvType) {
-		this.userSetEnvType = userSetEnvType;
+        this.userSetEnvType = userSetEnvType;
+
+        if(this.tagsList == null) {
+            this.tagsList = new ArrayList<>();
+        }
+
```
gauravakto, line 330: need to fix that duplicated adding bug as well. See the cyborg branch.

https://github.com/akto-api-security/akto/blob/90ae0ade0a05d86e48577cd06c45196eb2f5f1f6/libs/dao/src/main/java/com/akto/dto/ApiCollection.java#L318

PR #2423, opened 2025-04-28
Feature/module heartbeat handling

shivam-rawat-akto:


libs/dao/src/main/java/com/akto/dto/monitoring/ModuleInfo.java
```diff
@@ -0,0 +1,72 @@
+package com.akto.dto.monitoring;
+
+import java.util.UUID;
+
+public class ModuleInfo {
+
+    private ModuleType moduleType;
+    private String currentVersion;
+    private String id;//UUID
+    private int startedTs;
```
Ark2307, line 10: use that lombok import for clean coding?

notshivansh (approved):

PR #2424, opened 2025-04-28
cyborg handling for mini-runtime heartbeat

shivam-rawat-akto:


libs/utils/src/main/java/com/akto/data_actor/DbLayer.java
```diff
@@ -122,6 +124,21 @@ public static void updateApiCollectionName(int vxlanId, String name) {
         );
     }
 
+    public static void updateModuleInfo(ModuleInfo moduleInfo) {
+        ModuleInfo moduleInfoFromDB = ModuleInfoDao.instance.findOne(Filters.eq(ModuleInfo.MODULE_TYPE, moduleInfo.getModuleType()));
```
notshivansh, line 128: findOne should be on _id, because it is UUID, and will always be unique.

notshivansh: Alternatively, we can remove the findOne and UpdateOne commands, and add a single updateOne with upsert: true, based on _id.

notshivansh (approved):

PR #2512, opened 2025-05-14
Feature/metrics mini runtime cyborg

shivam-rawat-akto:


libs/dao/src/main/java/com/akto/dao/metrics/MetricDataDao.java
```diff
@@ -0,0 +1,95 @@
...
+        boolean exists = false;
+        String dbName = Context.accountId.get()+"";
+        MongoDatabase db = clients[0].getDatabase(dbName);
+        for (String col: db.listCollectionNames()){
+            if (getCollName().equalsIgnoreCase(col)){
+                exists = true;
+                break;
+            }
+        };
+
```
Ark2307, line 34: we are moving away from capped collections, ttl exploring https://www.mongodb.com/docs/v7.0/core/index-ttl/

libs/dao/src/main/java/com/akto/dao/metrics/MetricDataDao.java
```diff
@@ -0,0 +1,95 @@
...
+        if (!exists) {
+            if (DbMode.allowCappedCollections()) {
+                db.createCollection(getCollName(), new CreateCollectionOptions().capped(true).maxDocuments(maxDocuments).sizeInBytes(sizeInBytes));
+            } else {
+                db.createCollection(getCollName());
+            }
+        }
+
+        String[] fieldNames = {Log.TIMESTAMP};
+        MCollection.createIndexIfAbsent(getDBName(), getCollName(), fieldNames,false);
```
Ark2307, line 44: create compound index also, metric and ts

libs/dao/src/main/java/com/akto/dao/metrics/MetricDataDao.java
shivam-rawat-akto: till the time we have not moved to ttl, 
lets use capped collection

libs/dao/src/main/java/com/akto/dao/metrics/MetricDataDao.java
shivam-rawat-akto: we don't need compound index, 

as we use timestamp only while querying

libs/dao/src/main/java/com/akto/dao/metrics/MetricDataDao.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public Class<MetricData> getClassT() {
+        return MetricData.class;
+    }
+
+    public List<MetricData> getMetricsForTimeRange(String metricId, int accountId, long startTime, long endTime) {
+        List<Bson> filters = new ArrayList<>();
+        filters.add(Filters.eq("metricId", metricId));
```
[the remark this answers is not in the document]
shivam-rawat-akto: removed it

Ark2307 (approved):

PR #2513, opened 2025-05-14
Feature/metrics for runtime mini runtime

shivam-rawat-akto:


libs/utils/src/main/java/com/akto/metrics/AllMetrics.java
```diff
@@ -75,8 +78,10 @@ public void init(LogDb module, boolean pgMetrics){
                 kafkaBytesConsumedRate, cyborgNewApiCount, cyborgTotalApiCount, deltaCatalogNewCount, deltaCatalogTotalCount,
                 cyborgApiPayloadSize, multipleSampleDataFetchLatency);
 
-        executorService.scheduleAtFixedRate(() -> {
+        AllMetrics _this = this;
+        executorService.scheduleWithFixedDelay(() -> {
             try {
+                Context.accountId.set(accountId);
```
Ark2307, line 84: redundant code

shivam-rawat-akto: its an executor which is spawning its own thread, 

for which we need a threadlocal accountId

Ark2307 (approved):

PR #2561, opened 2025-05-23
Feature/add versioning in apis

Ark2307:


apps/api-runtime/src/main/java/com/akto/runtime/APICatalogSync.java
```diff
@@ -1978,4 +2013,21 @@ public APICatalog getDelta(int apiCollectionId) {
...
+
+    public boolean isMergeUrlsOnVersions() {
+        return mergeUrlsOnVersions;
+    }
+
+    public void setMergeUrlsOnVersions(boolean mergeUrlsOnVersions) {
+        this.mergeUrlsOnVersions = mergeUrlsOnVersions;
+    }
+
+    public boolean isSkipMergingOnKnownStaticURLsForVersionedApis() {
```
ayushaga14, line 2025: is or get?

apps/api-runtime/src/main/java/com/akto/runtime/APICatalogSync.java
```diff
@@ -73,6 +72,10 @@ public class APICatalogSync {
     public Map<SensitiveParamInfo, Boolean> sensitiveParamInfoBooleanMap;
     public static boolean mergeAsyncOutside = true;
     public BloomFilter<CharSequence> existingAPIsInDb = BloomFilter.create(Funnels.stringFunnel(Charsets.UTF_8), 1_000_000, 0.001 );
+    private boolean skipMergingOnKnownStaticURLsForVersionedApis = false;
```
ayushaga14, line 75: why is this false by default?

Ark2307: Coz for retrospective fix only we need that, not by default

apps/api-runtime/src/main/java/com/akto/runtime/APICatalogSync.java
Ark2307: getter of that boolean, auto-generated

ayushaga14 (approved):

PR #2568, opened 2025-05-26
Feature/support versions in urls mt

Ark2307:


apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/APICatalogSync.java
```diff
@@ -70,11 +70,13 @@ public class APICatalogSync {
     private static DataActor dataActor = DataActorFactory.fetchInstance();
     public static Set<MergedUrls> mergedUrls;
     private static final ClientLayer clientLayer = new ClientLayer();
-
+    public static final Pattern VERSION_PATTERN = Pattern.compile("\\bv([1-9][0-9]?|100)\\b");
```
ayushaga14, line 73: where is this getting used?

Ark2307: removed

PR #2632, opened 2025-06-05
added sse support for testing

krngrover94:


libs/utils/src/main/java/com/akto/testing/ApiExecutor.java
```diff
@@ -534,4 +540,127 @@ private static OriginalHttpResponse sendWithRequestBody(OriginalHttpRequest requ
...
+                        String dataLine = scanner.nextLine();
+                        if (dataLine.startsWith("data:")) {
+                            String data = dataLine.substring(5).trim();
+                            session.messages.add(data);
+                        }
+                    }
+                }
+            } catch (Exception ignored) {}
+        }).start();
+        return session;
```
notshivansh, line 612: generally we add a response.close, explicitly for normal requests. we check the same, if needed for sse request.

notshivansh (approved):

PR #2645, opened 2025-06-09
Add authentication

gauravakto:
### Why
Some customers don't want to run the traffic processor on their infra. For them we will have a hosted solution.

### What
- Add a auth layer

#### Assumptions
- We are using the same token as shown in UI for hybrid saas.

### Test cases
- [ ] No authentication required by default.
- [ ] Authentication required when AKTO_DI_AUTHENTICATE=true
- [ ] No auth header sent, 401
- [ ] Auth header incorrect, 401
- [ ] Success case

### DIFF
--- a/apps/data-ingestion-service/src/main/java/com/akto/action/IngestionAction.java
+++ b/apps/data-ingestion-service/src/main/java/com/akto/action/IngestionAction.java
@@ -50,11 +50,11 @@ public class IngestionAction extends ActionSupport {
     private static void printLogs(String msg) {
         MAX_INFO_PRINT--;
         if(MAX_INFO_PRINT > 0) {
-            loggerMaker.infoAndAddToDb(msg, LoggerMaker.LogDb.DATA_INGESTION);
+            loggerMaker.warnAndAddToDb(msg);
         }
 
         if(MAX_INFO_PRINT == 0) {
-            loggerMaker.infoAndAddToDb("Info log print limit reached.", LoggerMaker.LogDb.DATA_INGESTION);
+            loggerMaker.warnAndAddToDb("Debug log print limit reached.");
         }
     }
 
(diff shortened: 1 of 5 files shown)


apps/data-ingestion-service/src/main/java/com/akto/filters/AuthFilter.java
```diff
@@ -0,0 +1,88 @@
...
+
+         PublicKey publicKey = getPublicKey();
+         return Jwts.parserBuilder()
+                 .setSigningKey(publicKey)
+                 .build()
+                 .parseClaimsJws(jwsString);
+    }
+
+    private static PublicKey getPublicKey() throws NoSuchAlgorithmException, InvalidKeySpecException, IOException {
+       String rsaPublicKey = System.getenv("RSA_PUBLIC_KEY");
```
gauravakto, line 40: We will have to put the database abstractor public key as env variable. Currently it gets stored in common.configs as HYBRID_SAAS. @ayushaga14 I hope this is fine ?

ayushaga14: haa, just make sure the new container will not be accessible via public

apps/data-ingestion-service/src/main/java/com/akto/action/IngestionAction.java
```diff
@@ -80,5 +80,9 @@ public List<IngestDataBatch> getBatchData() {
     public void setBatchData(List<IngestDataBatch> batchData) {
         this.batchData = batchData;
     }
+
+    public String healthCheck() {
```
ayushaga14, line 87: should we return some generic response, for ex - {"ok":"ok"}

libs/dao/src/main/java/com/akto/dto/IngestDataBatch.java
```diff
@@ -17,118 +22,10 @@ public class IngestDataBatch {
...
-    public String getSource() {
-        return source;
-    }
-    public void setSource(String source) {
-        this.source = source;
-    }
+    String direction;
+    String process_id;
+    String socket_id;
+    String daemonset_id;
```
ayushaga14, line 28: are these new variables used anywhere?

gauravakto: runtime is pushing them into kafka, if we don't add then some downstream processing might get affected.

PR #2666, opened 2025-06-12
add mcp sync job to discover in mcp server

nayanakto:


libs/utils/src/main/java/com/akto/mcp/McpSchema.java
nayanakto, line 1: I have copy pasted this file from the official mcp java library.
It was written in java 17. I have converted it to java8.
To avoid writing the deserialization logic and lots of models for mcp json rpc, I have directly copy pasted it 😬 

The deserialization logic for tools/list response is working as expected.

notshivansh (approved):

PR #2676, opened 2025-06-16
add max conn and keep alive conn limit in http client init

ayushaga14:


apps/threat-detection/src/main/java/com/akto/threat/detection/tasks/SendMaliciousEventsToBackend.java
```diff
@@ -23,10 +24,17 @@ public class SendMaliciousEventsToBackend extends AbstractKafkaConsumerTask<byte
 
   private final CloseableHttpClient httpClient;
   private static final LoggerMaker logger = new LoggerMaker(SendMaliciousEventsToBackend.class, LogDb.THREAT_DETECTION);
+  private static final PoolingHttpClientConnectionManager connManager = new PoolingHttpClientConnectionManager();
+  
 
   public SendMaliciousEventsToBackend(KafkaConfig trafficConfig, String topic) {
     super(trafficConfig, topic);
-    this.httpClient = HttpClients.createDefault();
+    connManager.setMaxTotal(100);
```
gauravakto, line 32: pool exhausted then also the API would fail ? Can we change the httpClient.Do request excetption to errorAndDB so we know when error occurs.

gauravakto (approved):

gauravakto (approved):

gauravakto (approved):

PR #2694, opened 2025-06-18
MCP Tools Discovery - API Runtime

nayanakto:


libs/utils/src/main/java/com/akto/mcp/McpSchema.java
nayanakto, line 1: Have copy pasted it from mcp java sdk. Since it was written in java 17, I have converted it into java 8 compatible version.

The same version is present in mini runtime. I will leave up to you if you want to review it or not :)

libs/utils/src/main/java/com/akto/testing/ApiExecutor.java
```diff
@@ -661,21 +661,14 @@ private static SseSession openSseSession(String host, boolean debug) throws Exce
...
-                    int idx = endpoint.indexOf("sessionId=");
-                    if (idx != -1) {
-                        session.sessionId = endpoint.substring(idx + 10);
-                    }
+                    session.endpoint = dataLine.substring(5).trim();
                     break;
                 }
             }
         }
-        scanner.close();
```
nayanakto: closing this will close the sse stream and hence stop receiving response from mcp server for any request.
I have removed it to get response from sse stream. Once we get response from sse, the stream will be closed in the finally block.

Note: Mcp sends response in sse stream and not in the request that was sent from client.

notshivansh (approved):

PR #2700, opened 2025-06-19
Feature/common template

notshivansh:


libs/dao/src/main/java/com/akto/dao/test_editor/YamlTemplateDao.java
```diff
@@ -20,6 +20,18 @@ public class YamlTemplateDao extends AccountsContextDao<YamlTemplate> {
 
     public static final YamlTemplateDao instance = new YamlTemplateDao();
 
+    public Map<String, List<String>> fetchCommonWordListMap() {
+        YamlTemplate commonTemplate = CommonTemplateDao.instance.findOne(Filters.empty());
```
ayushaga14: cache this

notshivansh: fixed.

PR #2713, opened 2025-06-24
Add get LLM response API

nayanakto:


apps/database-abstractor/src/main/java/com/akto/action/LLMAction.java
```diff
@@ -0,0 +1,91 @@
...
+import org.json.JSONObject;
+
+@Getter
+@Setter
+public class LLMAction extends ActionSupport {
+
+    private static final LoggerMaker logger = new LoggerMaker(LLMAction.class, LogDb.DB_ABS);
+    private static final String JARVIS_ENDPOINT = "http://jarvis.internal.akto.io/api/generate";
+    private static final Gson gson = new Gson();
+    private static final OkHttpClient client = new OkHttpClient.Builder()
```
notshivansh: use corehttpclient. it has proxy implemented.

nayanakto: done

notshivansh (approved):

PR #2716, opened 2025-06-24
Feature/testing module credentials

aanchalakto:
adding API token retrieved through traffic in running tests


libs/dao/src/main/java/com/akto/dto/testing/TestRoles.java
```diff
@@ -82,33 +94,136 @@ public AuthMechanism findMatchingAuthMechanism(RawApi rawApi) {
             return findDefaultAuthMechanism();
         }
 
-        for(AuthWithCond authWithCond: this.getAuthWithCondList()) {
+        String deafaultAuthHeader = "";
+        List<AuthWithCond> authWithConds = this.getAuthWithCondList();
```
notshivansh: put this in try catch or add null checks for all fields , L98-L107

PR #2746, opened 2025-07-02
pass context to LLM without request

nayanakto:


libs/utils/src/main/java/com/akto/test_editor/Utils.java
```diff
@@ -1070,6 +1072,21 @@ public static String buildRequestIHttpFormat(RawApi rawApi) {
         return requestBuilder.toString();
     }
 
+    public static String buildEventStreamResponseIHttpFormat(OriginalHttpResponse response) {
+        if (response == null) {
+            return null;
+        }
```
krngrover94, line 1078: add response headers as well

krngrover94 (changes requested):

krngrover94 (approved):

PR #2747, opened 2025-07-02
add sample data auth to cyborg

ankush-jain-akto:


libs/dao/src/main/java/com/akto/DaoInit.java
```diff
@@ -74,6 +74,8 @@
 import com.mongodb.ReadPreference;
 import com.mongodb.client.MongoClients;
 
+import javassist.tools.reflect.Sample;
```
notshivansh, line 77: not needed

notshivansh (approved):

PR #2750, opened 2025-07-03
Add LLM Context - Mini

nayanakto:


libs/utils/src/main/java/com/akto/gpt/handlers/gpt_prompts/TestValidatorModifier.java
```diff
@@ -0,0 +1,39 @@
...
+            .append("- Determine whether it matches the behavior described by the following operation:\n")
+            .append("- The operation is: ")
+            .append(operation)
+            .append("\n----------------------------------------\n\n")
+            .append("Strict evaluation rules:\n")
+            .append("- If the response clearly reflects the operation, return a JSON object containing the correct operation key and value.\n")
+            .append("- Only use one of the approved JSON keys listed below:\n")
+            .append("  { \"contains_either\": [\"param1, param2\"] }\n")
+            .append("  { \"not_contains_either\": [\"param1, param2\"] }\n")
+            .append("  { \"contains_all\": [\"param1\"] }\n")
```
krngrover94, line 28: can we handle this in some other way?
if anytime new field gets added in yaml editor for validation
we need to change the prompt as well

libs/utils/src/main/java/com/akto/test_editor/Utils.java
```diff
@@ -995,6 +986,7 @@ public static String buildResponseIHttpFormat(RawApi rawApi) {
     }
 
     public final static String _MAGIC = "$magic";
+    public final static String MAGIC_CONTEXT = "$magic_context";
```
krngrover94, line 989: we can rename this in both the PRs

krngrover94 (changes requested):

libs/utils/src/main/java/com/akto/gpt/handlers/gpt_prompts/TestValidatorModifier.java
nayanakto: we discussed this :)

libs/utils/src/main/java/com/akto/test_editor/Utils.java
nayanakto: Would be great if there are any suggestions from your side!! would like to use the same. Meanwhile i will give it a thought on this

krngrover94 (approved):

notshivansh (approved):

PR #2770, opened 2025-07-07
breakdown context into chunks

nayanakto:


libs/utils/src/main/java/com/akto/gpt/handlers/gpt_prompts/PromptHandler.java
```diff
@@ -38,6 +48,111 @@ public BasicDBObject handle(BasicDBObject queryData) {
...
+                ? getContext(originalValue)
+                : originalValue;
+
+            if (StringUtils.isNotBlank(valueToSplit)) {
+                List<String> chunks = splitIntoChunks(valueToSplit);
+                if (chunks.size() > 1) {
+                    List<BasicDBObject> responses = new ArrayList<>();
+                    for (String chunk : chunks) {
+                        BasicDBObject queryDataCopy = (BasicDBObject) queryData.copy();
+                        queryDataCopy.put(key, key.equals(TestExecutorModifier._OPERATION)
```
krngrover94, line 69: why we need getContext and replaceContext
when we are splitting into chunks?

libs/utils/src/main/java/com/akto/gpt/handlers/gpt_prompts/PromptHandler.java
```diff
@@ -38,6 +48,111 @@ public BasicDBObject handle(BasicDBObject queryData) {
...
+                            existingArray.put(newArray.get(i));
+                        } catch (JSONException e) {
+                            logger.error("Error merging JSONArrays for key: " + key + ", value: " + newValue, e);
+                        }
+                    }
+                } else {
+                    aggregated.put(key, newValue);
+                }
+            }
+        }
```
krngrover94, line 151: Are we not adding this to testValidationModifier?

libs/utils/src/main/java/com/akto/gpt/handlers/gpt_prompts/PromptHandler.java
nayanakto: lets discuss this

libs/utils/src/main/java/com/akto/gpt/handlers/gpt_prompts/PromptHandler.java
nayanakto: PromptHandler is the base class so this entire logic will be automatically used for every classes who extends PromptHandler

krngrover94 (approved):

krngrover94 (approved):

PR #2787, opened 2025-07-11
modify string == checks to equals

ayushaga14:


ankush-jain-akto:
@ayushaga14 @gauravakto 

It should be `constant.equals(variable)`

PR #2789, opened 2025-07-14
Fix/kafka lag

gauravakto:


.github/workflows/prod.yml
```diff
@@ -74,20 +74,20 @@ jobs:
           IMAGE_TAG: latest
           IMAGE_TAG1: testruntime
           IMAGE_TAG2: local
-          IMAGE_TAG3: ${{ github.event.inputs.release_version }}_local
+          IMAGE_TAG3: ${{ github.event.inputs.release_version }}
```
gauravakto: Will revert this before merging

apps/mini-runtime/src/main/java/com/akto/hybrid_parsers/HttpCallParser.java
```diff
@@ -259,7 +295,16 @@ public static FILTER_TYPE isValidResponseParam(HttpResponseParams responseParam,
 
     public static Pair<HttpResponseParams,FILTER_TYPE> applyAdvancedFilters(HttpResponseParams responseParams, Map<String, List<ExecutorNode>> executorNodesMap,  Map<String,FilterConfig> filterMap){
         if (filterMap != null && !filterMap.isEmpty()) {
-            FILTER_TYPE filterType = isValidResponseParam(responseParams, filterMap, executorNodesMap);
+            FILTER_TYPE filterType = FILTER_TYPE.UNCHANGED; 
+
+            if (Context.accountId.get() != null && INPROCESS_ADVANCED_FILTERS_ACCOUNTS.contains(Context.accountId.get())){
```
ayushaga14: just verify once if Context.accountId here is being set correctly(should not be 1000000)

apps/mini-runtime/src/main/java/com/akto/hybrid_parsers/HttpCallParser.java
```diff
@@ -259,7 +295,16 @@ public static FILTER_TYPE isValidResponseParam(HttpResponseParams responseParam,
...
+
```
ayushaga14, line 301: Also i guess we can use actualAccountId variable, which is already defined in main.java

ayushaga14 (approved):

PR #2798, opened 2025-07-15
API for Cloudflare container to ingest traffic data

aanchalakto:


apps/mini-runtime-service/src/main/java/com/akto/action/IngestionAction.java
```diff
@@ -0,0 +1,183 @@
...
+                // Adding this if we are getting empty method from traffic connector
+                if(ACCOUNT_ID_TO_ADD_DEFAULT_DATA == 1745303931 && StringUtils.isEmpty(payload.getMethod())) {
+                    payload.setMethod("POST");
+                }
+
+                if(ACCOUNT_ID_TO_ADD_DEFAULT_DATA == 1745303931 && StringUtils.isEmpty(payload.getPath())) {
+                    payload.setPath("/");
+                }
+
+                HttpResponseParams responseParams = parseIngestDataBatchToHttpResponseParams(payload);
```
ayushaga14: keep function names readable

apps/mini-runtime-service/src/main/java/com/akto/action/IngestionAction.java
```diff
@@ -0,0 +1,183 @@
...
+        } catch (Exception e) {
+            loggerMaker.errorAndAddToDb("Error while inserting data to Kafka: " + e.getMessage(), LoggerMaker.LogDb.RUNTIME);
+            return Action.ERROR.toUpperCase();
+        }
+        return Action.SUCCESS.toUpperCase();
+    }
+
+    private final static ObjectMapper mapper = new ObjectMapper();
+
+    public static HttpResponseParams parseIngestDataBatchToHttpResponseParams(IngestDataBatch payloadData) throws JsonProcessingException {
```
ayushaga14: can we pass a major chunk of this function into some util file? which can be called by anyone later on

apps/mini-runtime-service/src/main/java/com/akto/action/IngestionAction.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        if(Utils.printDebugUrlLog(httpRequestParams.getURL())) {
+            printUrlDebugLog("Injecting K8 Pod Tags in Headers: " + tagsMap + " for URL: " + httpRequestParams.getURL());
+        }
+    }
+
+    public static int getAccountId() {
+        try {
+            String token = System.getenv("DATABASE_ABSTRACTOR_SERVICE_TOKEN");
+            DecodedJWT jwt = JWT.decode(token);
```
ayushaga14: i guess all these functions can be defined in some common util file, no need for duplication in mrs package

apps/mini-runtime-service/src/main/java/com/akto/listener/InitializerListener.java
```diff
@@ -0,0 +1,70 @@
...
+                    synchronized (IngestionAction.trafficDiscoveryQueue) {
+                        int size = IngestionAction.trafficDiscoveryQueue.size()> processingQueueThreshold ? processingQueueThreshold : IngestionAction.trafficDiscoveryQueue.size();
+                        for (int i = 0; i < size; i++) {
+                            toProcess.add(IngestionAction.trafficDiscoveryQueue.poll());
+                            loggerMaker.info("Processing element " + (i + 1) + " from traffic discovery queue");
+                        }
+                        // Remove processed elements
+                        for (int i = 0; i < size; i++) {
+                            IngestionAction.trafficDiscoveryQueue.remove(0);
+                        }
```
ayushaga14: can be done in the for loop before this

apps/mini-runtime-service/src/main/java/com/akto/listener/InitializerListener.java
```diff
@@ -0,0 +1,70 @@
...
+                    }
+                    com.akto.hybrid_runtime.Main.processData(toProcess);
```
ayushaga14, line 58: why not trigger this parallely? maybe we can keep 100 workers, and run this parallely

apps/mini-runtime-service/src/main/java/com/akto/listener/InitializerListener.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final LoggerMaker loggerMaker = new LoggerMaker(InitializerListener.class, LoggerMaker.LogDb.RUNTIME);
+    private static final ReentrantLock mutex = new ReentrantLock();
+    @Override
+    public void contextInitialized(ServletContextEvent sce) {
+        scheduler = Executors.newSingleThreadScheduledExecutor();
+        scheduler.scheduleAtFixedRate(() -> {
+            try {
+                mutex.lock();
+                long currentTime = System.currentTimeMillis();
+                if (IngestionAction.trafficDiscoveryQueue.size() > processingQueueThreshold
```
ayushaga14, line 44: can we also make sure that trafficDiscoveryQueue size doesn't blow up, like our processing speed is slower than data ingestion, than this can lead to process restarts due to OOM

apps/mini-runtime-service/src/main/test/java/com/akto/data-ingestion-service/ApiExecutorTest.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.*;
+
+import static org.junit.Assert.assertEquals;
+
+public class ApiExecutorTest {
+
+    @Test
+    public void kafkainteg() throws Exception {
+        KafkaUtils kafkaUtils = new KafkaUtils();
+        kafkaUtils.initKafkaProducer();
```
ayushaga14: is this file useful?

apps/mini-runtime-service/src/main/java/com/akto/utils/IngestDataBatch.java
```diff
@@ -0,0 +1,31 @@
...
+@lombok.Setter
+@lombok.AllArgsConstructor
+@lombok.NoArgsConstructor
+public class IngestDataBatch {
+
+    String path;
+    String requestHeaders;
+    String responseHeaders;
+    String method;
+    String requestPayload;
```
ayushaga14, line 12: can we use the same dto for both data ingestion service and mrs?

apps/mini-runtime-service/src/main/java/com/akto/listener/InitializerListener.java
```diff
@@ -0,0 +1,70 @@
...
+                        || (currentTime - lastProcessingTime > 5000 && !IngestionAction.trafficDiscoveryQueue.isEmpty())
+                        ) {
+                    loggerMaker.info("Processing traffic discovery queue with size: " + IngestionAction.trafficDiscoveryQueue.size());
+
+                    List<HttpResponseParams> toProcess = new ArrayList<>();
+                    synchronized (IngestionAction.trafficDiscoveryQueue) {
+                        int size = IngestionAction.trafficDiscoveryQueue.size()> processingQueueThreshold ? processingQueueThreshold : IngestionAction.trafficDiscoveryQueue.size();
```
ayushaga14, line 51: need to discuss this. 100 per sec might be too slow

ayushaga14 (changes requested):
resolve comments

apps/mini-runtime-service/src/main/java/com/akto/listener/InitializerListener.java
notshivansh: [ possible solution ] maybe we can add a limit to max size of queue, and make it configurable. and stop adding to queue, if it breaches that threshold. 
or something else.

apps/mini-runtime-service/src/main/java/com/akto/listener/InitializerListener.java
aanchalakto: Changed this to 100 per 10 ms and made it configurable value which can be updated in future dynamically

apps/mini-runtime-service/src/main/java/com/akto/utils/IngestDataBatch.java
aanchalakto: Yes, it is same but data ingestion service module is not present in mini-runtime branch, hence added this file

apps/mini-runtime-service/src/main/test/java/com/akto/data-ingestion-service/ApiExecutorTest.java
aanchalakto: Deleted

apps/mini-runtime-service/src/main/java/com/akto/listener/InitializerListener.java
aanchalakto: Increased the processing speed

apps/mini-runtime-service/src/main/java/com/akto/listener/InitializerListener.java
aanchalakto: done

notshivansh (approved):

ayushaga14 (approved):

PR #2871, opened 2025-07-29
Added functionality to add uningested api's into db

shivam-rawat-akto:

PR #2878, opened 2025-07-29
Feature/enable overage count dashboard

shivam-rawat-akto:


libs/dao/src/main/java/com/akto/dao/billing/UningestedApiOverageDao.java
notshivansh, line 1: same comments on this file from https://github.com/akto-api-security/akto/pull/2871

libs/dao/src/main/java/com/akto/dto/billing/UningesetedApiOverage.java
notshivansh, line 1: Same comments on this file from https://github.com/akto-api-security/akto/pull/2871

notshivansh, line 1: Register this in Daoinit.

PR #2896, opened 2025-07-30
Ingestion stop reproduce

gauravakto:


libs/utils/src/main/java/com/akto/runtime/parser/KafkaProducerExample.java
gauravakto, line 1: Ignore this. It's just data dumping into kafka

PR #2931, opened 2025-08-05
handle gzip encoded

notshivansh:


libs/utils/src/main/java/com/akto/runtime/parser/SampleParser.java
```diff
@@ -104,6 +108,39 @@ public static HttpResponseParams parseSampleMessage(String message) throws Excep
 
     }
 
+    private static final String CONTENT_ENCODING_HEADER = "content-encoding";
+    private static final String _GZIP = "gzip";
+
+    private static List<String> getHeaderIgnoreCase(Map<String, List<String>> headers, String headerName) {
+        for (Map.Entry<String, List<String>> entry : headers.entrySet()) {
+            if (entry.getKey() != null && entry.getKey().equalsIgnoreCase(headerName)) {
+                return entry.getValue();
```
Ark2307, line 117: it is a map, why loop, just get value?

notshivansh: we're not sure on the case of the header.
it could be content-encoding, Content-Encoding, content-Encoding, Content-encoding or any other permutation.

ayushaga14 (approved):

PR #2946, opened 2025-08-08
code optimization done

aanchalakto:


apps/dashboard/src/main/java/com/akto/action/observe/InventoryAction.java
```diff
@@ -1159,40 +1160,90 @@ public String undoDemergedApis() {
...
     public String fetchNotTestedAPICount() {
         Bson filterQ = UsageMetricCalculator.excludeDemosAndDeactivated(ApiInfo.ID_API_COLLECTION_ID);
-
         Bson filter = Filters.and(
                 filterQ,
                 Filters.exists(ApiInfo.LAST_TESTED, false)
         );
 
+        int totalCount = (int) ApiInfoDao.instance.count(filter);
+        int numThreads = Math.max(1, Runtime.getRuntime().availableProcessors());
```
notshivansh: we don't need parallelization at both frontend and backend.
Implement it at one place only. Frontend should be good for the same. 
Moreover, when implementing in backend, this way the behaviour would change from machine to machine. we would not want that

notshivansh:
Please try to reuse code. Make utility functions and DRY.

notshivansh (approved):

PR #2996, opened 2025-08-14
Mcp audit

aanchalakto:


apps/dashboard/src/main/java/com/akto/action/MCPScanAction.java
```diff
@@ -54,6 +56,14 @@ public String initiateMCPScan() {
                 loggerMaker.info("Creating ApiCollection for host: " + hostName, LogDb.DASHBOARD);  
                 createdCollection = new ApiCollection(collectionId, hostName, Context.now(), new HashSet<>(), hostName, 0, false, true, sseEndpoint);
                 ApiCollectionsDao.instance.insertOne(createdCollection);
+
+                //New MCP server detected, audit it
+                McpAuditInfo auditInfo = new McpAuditInfo();
+                auditInfo.setMarkedBy("System");
+                auditInfo.setResourceName("MCP Server detected: " + hostName);
+                auditInfo.setType("MCP Server");
```
Ark2307: use constructor for creating object

libs/utils/src/main/java/com/akto/mcp/McpRequestResponseUtils.java
```diff
@@ -82,12 +84,30 @@ private static void handleMcpMethodCall(McpJsonRpcModel mcpJsonRpcModel, String
             case McpSchema.METHOD_TOOLS_CALL:
                 if (params != null && StringUtils.isNotBlank(params.getName())) {
                     url = HttpResponseParams.addPathParamToUrl(url, params.getName());
+
+                    // Log the tool detection in audit info
+                    String toolName = params.getName();
+                    McpAuditInfo auditInfo = new McpAuditInfo();
+                    auditInfo.setMarkedBy("System");
+                    auditInfo.setResourceName("Tool detected: " + toolName+" " + url);
+                    auditInfo.setType("Tool");
```
Ark2307: create object above and set resource type according to switch case

libs/utils/src/main/java/com/akto/mcp/McpRequestResponseUtils.java
```diff
@@ -82,12 +84,30 @@ private static void handleMcpMethodCall(McpJsonRpcModel mcpJsonRpcModel, String
             case McpSchema.METHOD_TOOLS_CALL:
                 if (params != null && StringUtils.isNotBlank(params.getName())) {
                     url = HttpResponseParams.addPathParamToUrl(url, params.getName());
+
+                    // Log the tool detection in audit info
+                    String toolName = params.getName();
+                    McpAuditInfo auditInfo = new McpAuditInfo();
+                    auditInfo.setMarkedBy("System");
```
Ark2307: marked by is the field for approving/rejecting and not detecting

apps/dashboard/src/main/java/com/akto/action/MCPScanAction.java
```diff
@@ -54,6 +56,14 @@ public String initiateMCPScan() {
...
+                auditInfo.setLastDetected((int) (System.currentTimeMillis()));
+                McpAuditInfoDao.instance.insertOne(auditInfo);
```
nayanakto: In runtime we create a collection, whenever mcp server is detected for the first time. you need to add changes there as well.

Lets sync up if needed

apps/dashboard/src/main/java/com/akto/action/MCPScanAction.java
```diff
@@ -54,6 +56,10 @@ public String initiateMCPScan() {
                 loggerMaker.info("Creating ApiCollection for host: " + hostName, LogDb.DASHBOARD);  
                 createdCollection = new ApiCollection(collectionId, hostName, Context.now(), new HashSet<>(), hostName, 0, false, true, sseEndpoint);
                 ApiCollectionsDao.instance.insertOne(createdCollection);
+
+                //New MCP server detected, audit it
+                McpAuditInfo auditInfo = new McpAuditInfo(hostName, "MCP server", (int) (System.currentTimeMillis()), collectionId);
```
Ark2307: use tag from constants.java and time as context.now()

apps/dashboard/src/main/java/com/akto/action/MCPScanAction.java
```diff
@@ -54,6 +58,18 @@ public String initiateMCPScan() {
                 loggerMaker.info("Creating ApiCollection for host: " + hostName, LogDb.DASHBOARD);  
                 createdCollection = new ApiCollection(collectionId, hostName, Context.now(), new HashSet<>(), hostName, 0, false, true, sseEndpoint);
                 ApiCollectionsDao.instance.insertOne(createdCollection);
+
+                try {
+                    //New MCP server detected, audit it
+                    McpAuditInfo auditInfo = new McpAuditInfo(
+                            Context.now(), "", AKTO_MCP_SERVER_TAG , 0,
```
Ark2307, line 65: add check here for identifying mcp collection or not

aanchalakto: I think it is MCP scan and by default the collections here would be mcp collection. @nayanakto can you throw some light and let me know if my understanding is correct

nayanakto: This is the correct place as this API is fired from dashboard - MCP Import

PR #3140, opened 2025-09-09
Fetch paginated api rate limits

gauravakto:


libs/dao/src/main/java/com/akto/dao/ApiInfoDao.java
```diff
@@ -76,6 +78,11 @@ public void createIndicesIfAbsent() {
 
         MCollection.createIndexIfAbsent(getDBName(), getCollName(),
                 new String[] { ApiInfo.RISK_SCORE, ApiInfo.ID_API_COLLECTION_ID }, false);
+        
+        // Create sparse index for rate limits pagination
+        IndexOptions sparseIndex = new IndexOptions().sparse(true);
+        Bson rateLimitsIndex = Indexes.ascending("rateLimits", "rateLimitConfidence", "_id");
+        createIndexIfAbsent(getDBName(), getCollName(), rateLimitsIndex, sparseIndex.name("rateLimits_confidence_id_sparse"));
```
gauravakto, line 85: need to test this, on server startup the index was not getting created automatically.

libs/dao/src/main/java/com/akto/dto/ApiInfo.java
```diff
@@ -48,6 +48,12 @@ public class ApiInfo {
     public static final String LAST_CALCULATED_TIME = "lastCalculatedTime";
     private int lastCalculatedTime;
 
+    public static final String RATELIMITS = "rateLimits";
+    private Map<String, Integer> rateLimits;
+    
+    public static final String RATE_LIMIT_CONFIDENCE = "rateLimitConfidence";
+    private float rateLimitConfidence;
+
```
gauravakto, line 56: New fields added in APIInfo, address the comment at the top of APIInfo.java file to add these in relevant places.

gauravakto: Added handling in merge method to pick the max values.


@notshivansh Do I need to change something in AktoPolicy too ??? (the comment at the top of file says so)

notshivansh (approved):

PR #3154, opened 2025-09-10
feature: add dynamic api error calculation

harshithb3304:
…n logic

### DIFF
--- a/apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
+++ b/apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
@@ -77,6 +77,7 @@ import java.util.*;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 
 public class DbAction extends ActionSupport {
     static final ScheduledExecutorService service = Executors.newSingleThreadScheduledExecutor();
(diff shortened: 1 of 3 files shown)


apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -1887,7 +1918,12 @@ public String insertTestingRunResultSummary() {
 
     public String insertTestingRunResults() {
         try {
-
+                if (Context.accountId.get() == null || Context.accountId.get() == 0) {
```
notshivansh: remove this. context.account is always set

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -1932,6 +1971,59 @@ public String insertTestingRunResults() {
                 testingRunResult.setTestRunResultSummaryId(id);
             }
 
+            try {
```
notshivansh: we need the API to be fast. 
make a scheduleExecutor with fixed threads [ say 10 ]. 
let the actual update happen [ without apiErrors ]. 
submit a job to executor which runs the regex.
if apiErrors are > 0 , issue an update, else skip updates altogether. 
this will help in better find query performance, because we would be able to use the $exists paradigm.

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -1932,7 +1973,83 @@ public String insertTestingRunResults() {
...
+            // insert immediately without apiErrors to keep the insert fast and keep the field absent when not needed
             DbLayer.insertTestingRunResults(testingRunResult);
+
+            // submit async job to compute apiErrors and update only if any error flag > 0
+            final TestingRunResult trrForAsync = testingRunResult;
+            final int capturedAccountId = (Context.accountId.get() != null) ? Context.accountId.get() : 0;
+            apiErrorsService.submit(() -> {
+                if (capturedAccountId != 0) {
+                    Context.accountId.set(capturedAccountId);
+                }
```
notshivansh: don't run the thread altogether, if accountId is 0

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -1932,7 +1973,83 @@ public String insertTestingRunResults() {
...
+
+                    apiErrors.put("cloudflare", cloudflareErrors);
+                    apiErrors.put("429", rateLimit429);
+                    apiErrors.put("5xx", server5xx);
+
+                    // only update if any error flag is > 0 (keeps field absent when not needed, allows $exists queries)
+                    if (cloudflareErrors + rateLimit429 + server5xx > 0) {
+                        try {
+                            // set apiErrors field on the existing document
+                            org.bson.conversions.Bson filter = Filters.eq("_id", trrForAsync.getId());
```
notshivansh: always use string literals, for all string fields.
i.e. instead of `_id`, use `Constants._ID` and define it somewhere [ I think it's already defined somewhere ].

PR #3169, opened 2025-09-12
Feature/testing visibility

harshithb3304:


apps/dashboard/src/main/java/com/akto/action/testing/TestResultsStatsAction.java
```diff
@@ -258,12 +257,16 @@ private int getCountByPattern(ObjectId testingRunResultSummaryId, String regex)
...
+            if (key == null) return false;
+
+            Bson filter = Filters.and(
+                    Filters.eq("testRunResultSummaryId", testingRunResultSummaryId),
+                    Filters.eq("vulnerable", false),
+                    Filters.exists("apiErrors." + key, true)
+            );
+
+            long count = TestingRunResultDao.instance.getRawCollection().countDocuments(filter);
+            return count > 0;
```
notshivansh: replace this with findOne.
The point here is to check if it has any `apiErrors` key, if it has one, it likely will have more as well.
Also, we don't need to check for specific keys, `"apiErrors." + key` checking just the top level opject `apiErrors` would work

apps/dashboard/src/main/java/com/akto/action/testing/TestResultsStatsAction.java
```diff
@@ -272,26 +275,25 @@ private boolean hasApiErrors(ObjectId testingRunResultSummaryId) {
...
                         Filters.eq("vulnerable", false),
-                        Filters.exists("apiErrors." + key, true))));
-
+                        Filters.exists("apiErrors." + key, true)
+                )));
         pipeline.add(Aggregates.group(null, Accumulators.sum("count", "$apiErrors." + key)));
 
-        MongoCursor<BasicDBObject> cursor = TestingRunResultDao.instance.getMCollection()
-                .aggregate(pipeline, BasicDBObject.class).cursor();
+        MongoCursor<Document> cursor = TestingRunResultDao.instance.getRawCollection()
```
notshivansh: why are we using rawCollection ?

libs/dao/src/main/java/com/akto/dao/testing/TestingRunResultDao.java
```diff
@@ -258,6 +258,9 @@ public void createIndicesIfAbsent() {
         fieldNames = new String[]{TestingRunResult.REQUIRES_CONFIG};
         MCollection.createIndexIfAbsent(getDBName(), getCollName(), fieldNames, false);
 
+        // Index on apiErrors map to speed up existence and aggregation queries
+        MCollection.createIndexIfAbsent(getDBName(), getCollName(), new String[]{TestingRunResult.API_ERRORS}, false);
```
notshivansh: make this index composite. i.e. in the future we should be able to extend it.
and make it more useful [ brings the least no. of documents required ]
I think if we add `summaryId` and `vulnerable` field to it, along as well, should be good.
check for usage of index in all cases.

libs/dao/src/main/java/com/akto/dto/testing/TestingRunResult.java
```diff
@@ -47,13 +47,22 @@ public class TestingRunResult implements Comparable<TestingRunResult> {
...
     public static final String IS_IGNORED_RESULT = "isIgnoredResult";
-    private boolean isIgnoredResult ;
+    private boolean isIgnoredResult;
+    @BsonIgnore
+    private String testRunResultSummaryHexId;
+    @BsonIgnore
+    private List<TestResult> singleTestResults;
+    @BsonIgnore
+    private List<MultiExecTestResult> multiExecTestResults;
+
```
notshivansh: why do we need these ?

PR #3185, opened 2025-09-17
Mini runtime mcp recon

aanchalakto:


apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/Main.java
```diff
@@ -483,6 +483,19 @@ public void run() {
...
+        scheduler.scheduleAtFixedRate(() -> {
+            Context.accountId.set(DataActor.actualAccountId);
+            try {
+                loggerMaker.infoAndAddToDb("Executing MCP Recon Sync job");
+                McpReconSyncJobExecutor.INSTANCE.runJob();
+                loggerMaker.infoAndAddToDb("Finished executing MCP Recon Sync job");
+            } catch (Exception e) {
+                loggerMaker.errorAndAddToDb(e, "Error while executing MCP Recon Sync Job");
+            }
+        }, 0, 2, TimeUnit.MINUTES);
```
notshivansh: 2 minutes, isn't very frequent ?

libs/dao/src/main/java/com/akto/DaoInit.java
```diff
@@ -383,6 +383,9 @@ public static void createIndices() {
         CodeAnalysisCollectionDao.instance.createIndicesIfAbsent();
         CodeAnalysisApiInfoDao.instance.createIndicesIfAbsent();
         JobsDao.instance.createIndicesIfAbsent();
+        McpReconRequestDao.instance.createIndicesIfAbsent();
+        McpReconResultDao.instance.createIndicesIfAbsent();
+
```
notshivansh, line 391: btw, these also need to be added in the dashboard pr, https://github.com/akto-api-security/akto/pull/3190 , and in the dashboard jobs branch.

apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/McpReconScanner.java
```diff
@@ -0,0 +1,797 @@
...
+            
+        this.httpClient = HttpClientBuilder.create()
+            .setConnectionManager(connectionManager)
+            .setDefaultRequestConfig(requestConfig)
+            .setConnectionReuseStrategy((response, context) -> true)  // Always reuse connections
+            .setKeepAliveStrategy((response, context) -> 30000)  // Keep connections alive for 30s
+            .setMaxConnPerRoute(50)
+            .setMaxConnTotal(maxConcurrent * 2)
+            .build();
+    }
```
notshivansh: please use CoreHttpClient to make any http connection.
It has proxies, tls settings etc. implemented at a global level.
The pooling we can add on top of it.

notshivansh: So, we use OkHTTP library for all connections.
Alternatively, you can directly use ApiExecutor.java's methods to make http calls as well.

aanchalakto: done

notshivansh (approved):

PR #3229, opened 2025-09-24
deployment configs added

harshithb3304:


apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -3959,4 +3962,195 @@ public void setTestingRunResultId(String testingRunResultId) {
...
+    // Deployment configuration fields
+    private List<DeploymentConfig> deploymentConfigs;
+    private String deploymentId;
+    private String deploymentName;
+    private String deploymentType;
+    private List<EnvVariable> envVars;
+    private String envKey;
+    private String envValue;
+    private boolean editable;
+
```
notshivansh, line 4018: use getter/setter annotation. remove explicit declarations for the same
I think getter are for response and setter are for request. [ or vice versa ].
check and use accordinly.

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -3959,4 +3962,195 @@ public void setTestingRunResultId(String testingRunResultId) {
...
+            return SUCCESS;
+        } catch (Exception e) {
+            loggerMaker.errorAndAddToDb(e, "Error in updateEnvVariable " + e.toString());
+            return ERROR;
+        }
+    }
+
+    public String removeEnvVariable() {
+        try {
+            if (deploymentId == null || envKey == null) {
```
notshivansh: why would we need these APIs in cyborg ?
only mini-runtime/testing i.e. client modules talk to cyborg.
please add only APIs which we need.

PR #3255, opened 2025-09-29
remove nin query

ayushaga14:


libs/utils/src/main/java/com/akto/usage/UsageMetricCalculator.java
```diff
@@ -185,29 +185,71 @@ public static int calculateCustomTests(UsageMetric usageMetric) {
...
+        return collectionIds;
+    }
+
+    /**
+     * Get valid collection IDs (all collections excluding demos and deactivated)
+     * This avoids using $nin queries in favor of $in queries
+     */
+    private static Set<Integer> getValidCollectionIds() {
+        // Get all API collections
+        List<ApiCollection> allCollections = ApiCollectionsDao.instance.findAll(Filters.empty());
```
Ark2307: get only ids and that too not of api_group type

libs/utils/src/main/java/com/akto/usage/UsageMetricCalculator.java
```diff
@@ -185,29 +185,71 @@ public static int calculateCustomTests(UsageMetric usageMetric) {
...
+        
+        // Get demo and deactivated collection IDs
+        Set<Integer> demosAndDeactivated = getDemosAndDeactivated();
```
Ark2307: don't call this, instead apply filter to get only active and remove demo, this is called from cache and may call api again

libs/utils/src/main/java/com/akto/usage/UsageMetricCalculator.java
```diff
@@ -185,29 +185,71 @@ public static int calculateCustomTests(UsageMetric usageMetric) {
     public static int calculateTestRuns(UsageMetric usageMetric) {
         int measureEpoch = usageMetric.getMeasureEpoch();
 
-        Bson demoAndDeactivatedCollFilter = excludeDemosAndDeactivated(TestingRunResult.API_INFO_KEY + "." + ApiInfo.ApiInfoKey.API_COLLECTION_ID);
+        // Get valid collection IDs (excluding demos and deactivated collections)
+        Set<Integer> validCollectionIds = getValidCollectionIds();
 
+        // Apply RBAC filtering to get user-allowed collection IDs
+        Set<Integer> rbacFilteredCollectionIds = applyRbacFiltering(validCollectionIds);
```
Ark2307: dont call this here

PR #3263, opened 2025-09-30
add stigg check

ayushaga14:


apps/threat-detection/src/main/java/com/akto/threat/detection/Main.java
```diff
@@ -34,7 +39,25 @@ public class Main {
   private static final DataActor dataActor = DataActorFactory.fetchInstance();
 
   public static void main(String[] args) throws Exception {
-    
+
+    while (true) {
+
+      int accountId = ClientActor.getAccountId();
+
+      Organization organization = dataActor.fetchOrganization(accountId);
```
Ark2307, line 47: sometimes organization maybe null, add null check

gauravakto, line 47: organization null check.

and featureWiseAllowed isEmpty check

PR #3281, opened 2025-10-03
Imperva Import Feature - UI

nayanakto:


libs/utils/src/main/java/com/akto/imperva/parser/ImpervaSchemaParser.java
```diff
@@ -200,10 +200,11 @@ private static boolean isValidContentType(String contentType) {
      * Checks if content-type is XML or SOAP based.
      */
     private static boolean isXmlContentType(String contentType) {
-        return contentType != null && (
-            contentType.contains("xml") ||
-            contentType.contains("soap")
-        );
+        // return contentType != null && (
```
Ark2307: remove this if not need, also we have a function which checks this

PR #3286, opened 2025-10-06
MCP Recon code reusability added for tool, resource discovery, issue …

aanchalakto:
…fixed and capability added

### DIFF
--- a/apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/McpToolsSyncJobExecutor.java
+++ b/apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/McpToolsSyncJobExecutor.java
@@ -161,34 +161,7 @@ public class McpToolsSyncJobExecutor {
         List<HttpResponseParams> responseParamsList = new ArrayList<>();
         if(isRecon) {
             host = mcpServer.getIp() + ":" + mcpServer.getPort();
-
-            try {
-                int id = 1;
-                String toolsCallRequestHeaders = buildHeaders(host);
-                for (McpSchema.Tool tool : mcpServer.getTools()) {
-                    McpSchema.JSONRPCRequest request = new McpSchema.JSONRPCRequest(
-                            McpSchema.JSONRPC_VERSION,
-                            McpSchema.METHOD_TOOLS_CALL,
-                            id++,
-                            new McpSchema.CallToolRequest(tool.getName(), McpToolsSyncJobExecutor.generateExampleArguments(tool.getInputSchema()))
-                    );
-
-                    HttpResponseParams toolsCallHttpResponseParams = McpToolsSyncJobExecutor.convertToAktoFormat(0,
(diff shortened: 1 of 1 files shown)


apps/mini-runtime/src/main/java/com/akto/hybrid_parsers/HttpCallParser.java
```diff
@@ -507,13 +508,16 @@ private static void sendTrafficMetricsToTelemetry(BasicDBObject metricsData) {
         }
     }
 
-    public static boolean useHostCondition(String hostName, HttpResponseParams.Source source) {
+    public static boolean useHostCondition(String hostName, HttpResponseParams.Source source, boolean isRecon) {
```
gauravakto: HttpResponseParams is already available everywhere. Instead of passing a new variable in all the places. 

We can simply create a new source  HttpResponseParams.Source.MCP_RECON

Your detection mechanism can set the source and here you can check for it.


This can in future be used by others or for fitlering( I think we do store the source somewhere )

aanchalakto: done

gauravakto: HttpResopnse param is a very widely used class, please do regression testing explicitly before deploying

Pipeline
Traffic Collector(daemonset) -> kafka -> Mini runtime -> Dashboard UI

1. Make sure traffic ingestion is happening
2. Make sure changes are backwards compatible.

gauravakto: @nayanakto @aanchalakto I have not checked the rest of the code. Please review.

I just checked the hostCondition thing.

apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/McpToolsSyncJobExecutor.java
```diff
@@ -153,123 +154,190 @@ private List<HttpResponseParams> initializeMcpServerCapabilities(ApiCollection a
...
-                    if (normalizedSampleDataSet.contains(McpSchema.METHOD_TOOLS_CALL + "/" + tool.getName())) {
-                        logger.debug("Skipping tool {} as it is already present in the db.", tool.getName());
-                        continue;
-                    }
-                    JSONRPCRequest request = new JSONRPCRequest(
-                        McpSchema.JSONRPC_VERSION,
-                        McpSchema.METHOD_TOOLS_CALL,
-                        id++,
-                        new CallToolRequest(tool.getName(), generateExampleArguments(tool.getInputSchema()))
+                for (McpSchema.Tool tool : mcpServer.getTools()) {
```
nayanakto: lets extract out this method as it is used twice.
you can refer to the extracted method from your recon class

PR #3289, opened 2025-10-07
guardrail show threats

rakshak-akto:


apps/threat-detection-backend/src/main/java/com/akto/threat/backend/db/MaliciousEventModel.java
```diff
@@ -38,6 +39,38 @@ public enum Status {
     IGNORED
   }
 
+  public enum Label {
```
gauravakto, line 42: just use normal enum like above.

apps/threat-detection-backend/src/main/java/com/akto/threat/backend/service/MaliciousEventService.java
```diff
@@ -78,6 +78,7 @@ public void recordMaliciousEvent(String accountId, RecordMaliciousEventRequest r
             .setType(evt.getType())
             .setMetadata(evt.getMetadata().toString())
             .setSuccessfulExploit(evt.getSuccessfulExploit())
+            .setLabel(evt.getLabel())
```
gauravakto: Threat Client -> MaliciousDetectorTask -> buildSampleMaliciousRequest -> from here also we need to send the label as THREAT now.

rakshak-akto: https://github.com/akto-api-security/akto/pull/3300

PR #3304, opened 2025-10-09
supporting streamable http in mcp

shivam-rawat-akto:


libs/utils/src/main/java/com/akto/testing/ApiExecutor.java
krngrover94, line 764: here we are checking for header

libs/utils/src/main/java/com/akto/testing/ApiExecutor.java
```diff
@@ -755,6 +755,11 @@ private static boolean shouldInitiateSSEStream(OriginalHttpRequest request) {
             return true;
         }
 
+        // Return false for streamable HTTP requests for MCP
+        if ("/mcp".equals(request.getPath())) {
+            return false;
```
krngrover94: what if sse server has /mcp path?
And we are already checking for event-stream headers which usually exist in case of sse request?
Have you find any case where it is failing?

PR #3309, opened 2025-10-09
Updated McpAudit dto

aanchalakto:


libs/dao/src/main/java/com/akto/dto/McpAuditInfo.java
```diff
@@ -27,6 +28,12 @@ public class McpAuditInfo {
     private Set<ApiInfo.ApiAccessType> apiAccessTypes;
     private int hostCollectionId;
 
+    // Conditional approval fields
+    private Map<String, Object> approvalConditions;
```
rakshak-akto, line 32: remove the constructor since its not used.

PR #3339, opened 2025-10-15
added new LogDb cyborg

harshithb3304:


libs/utils/src/main/java/com/akto/data_actor/DbLayer.java
```diff
@@ -681,6 +681,10 @@ public static void insertProtectionLog(Log log) {
         ProtectionLogsDao.instance.insertOne(log);
     }
 
+    public static void insertCyborgLog(Log log) {
+        com.akto.dao.CyborgLogsDao.instance.insertOne(log);
```
gauravakto: import at top

harshithb3304: done

PR #3388, opened 2025-10-21
Fixing prompt handler using azure open ai

Ark2307:


krngrover94 (approved):

libs/utils/src/main/java/com/akto/gpt/handlers/gpt_prompts/TestExecutorModifier.java
```diff
@@ -7,7 +7,9 @@
 
 import com.mongodb.BasicDBObject;
 
-public class TestExecutorModifier extends PromptHandler {
+public class TestExecutorModifier extends AzureOpenAIPromptHandler {
+
```
krngrover94, line 11: we have 2 more magic filters other than this
TestFilterModifier
TestValidatorModifier

PR #3389, opened 2025-10-21
added accountId to each log

harshithb3304:


libs/utils/src/main/java/com/akto/log/LoggerMaker.java
```diff
@@ -331,19 +331,27 @@ public List<Log> fetchLogRecords(int logFetchStartTime, int logFetchEndTime, Log
...
     public void info(String message, Object... vars) {
-        logger.info(message, vars);
+        String accountId = Context.accountId.get() != null ? Context.accountId.get().toString() : "NA";
+        String formattedMessage = "accountId: " + accountId + ", " + message;
+        logger.info(formattedMessage, vars);
     }
 
     public void error(String errorMessage, Object... vars) {
-        logger.error(errorMessage, vars);
+        String accountId = Context.accountId.get() != null ? Context.accountId.get().toString() : "NA";
```
gauravakto: create a common function


getFormattedMessage use it everyhwere

libs/utils/src/main/java/com/akto/log/LoggerMaker.java
```diff
@@ -157,16 +157,17 @@ protected static void sendToSlack(String err) {
     }
 
     protected String basicError(String err, LogDb db) {
-        if(Context.accountId.get() != null){
-            err = String.format("%s\nAccount id: %d", err, Context.accountId.get());
+        if (Context.accountId.get() != null) {
+            err = String.format("%s\naccountId: %d", err, Context.accountId.get());
         }
```
gauravakto: we don't need this? formattedMessage will anyways add the accountID

libs/utils/src/main/java/com/akto/log/LoggerMaker.java
```diff
@@ -331,19 +330,27 @@ public List<Log> fetchLogRecords(int logFetchStartTime, int logFetchEndTime, Log
...
-        logger.warn(message, vars);
+        logger.warn(getFormattedMessage(message), vars);
+    }
+
+    /**
+     * Helper to prefix the message with accountId consistently.
+     */
+    private String getFormattedMessage(String message) {
+        String accountId = Context.accountId.get() != null ? Context.accountId.get().toString() : "NA";
+        return "accountId: " + accountId + ", " + message;
```
gauravakto: For each log line we need to make it parseable by code/regex. For that we must have specfic format. Let's keep this for now.

`return "accountId: " + accountId + " message:  " + message;

PR #3397, opened 2025-10-23
add locale detection

ayushaga14:


libs/utils/src/main/java/com/akto/runtime/RuntimeUtil.java
```diff
@@ -78,5 +82,33 @@ public static boolean isValidVersionToken(String token){
...
+        String[] parts = token.split("-");
+
+        // Check language code (2 or 3 letters)
+        if (parts.length == 1) {
+            // Just language code: "en", "ja"
+            return VALID_LANGUAGE_CODES.contains(parts[0].toLowerCase());
+        } else if (parts.length == 2) {
+            // Language + country: "en-US", "ja-JP"
+            String language = parts[0].toLowerCase();
+            String country = parts[1].toUpperCase();
```
gauravakto, line 104: check if getISOLanguages()) and getISOCountries() also return always in lowe and upper case

gauravakto (dismissed):

libs/utils/src/main/java/com/akto/runtime/RuntimeUtil.java
ayushaga14: yes, checked

PR #3428, opened 2025-10-28
avoid merging if demerging applied before

ayushaga14:


apps/database-abstractor/src/main/java/com/akto/merging/MergingLogic.java
```diff
@@ -33,7 +35,37 @@ public class MergingLogic {
...
+            loggerMaker.infoAndAddToDb("Loaded " + mergedUrls.size() + " demerged URLs", LogDb.DB_ABS);
+        } catch (Exception e) {
+            loggerMaker.errorAndAddToDb("Error loading merged URLs: " + e.getMessage(), LogDb.DB_ABS);
+            mergedUrls = new HashSet<>();
+        }
+    }
+
+    private static boolean isDemergedUrl(URLTemplate urlTemplate, int apiCollectionId) {
+        try {
+            for (MergedUrls mergedUrl : mergedUrls) {
```
gauravakto, line 52: nitpick... create an equals method on MergedUrls?? So we can directly do set.contains() ??

gauravakto (approved):

PR #3456, opened 2025-10-31
remove tryForCollectionName function

ayushaga14:


gauravakto (dismissed):

notshivansh (approved):

libs/dao/src/main/java/com/akto/dto/OriginalHttpRequest.java
```diff
@@ -242,17 +246,60 @@ public static Map<String,List<String>> buildHeadersMap(Map json, String key) {
...
-            values.add(headersFromRequest.get(k).toString());
-            headers.put(k.toString().toLowerCase(),values);
+        if (headersString == null || headersString.isEmpty()) return headers;
+
+        try (JsonParser p = JSON_FACTORY.createParser(headersString)) {
+            // Expect a single JSON object
+            if (p.nextToken() != JsonToken.START_OBJECT) return headers;
+
+            String currentKey = null;
+            while (p.nextToken() != JsonToken.END_OBJECT) {
```
gauravakto, line 257: this is the same alibaba library... if closing } is not found then infinite loop.

PR #3469, opened 2025-11-03
Add MCP registry status handling and UI components

krngrover94:
- Implemented `McpRegistryUtils` to check MCP server availability in the registry.
- Added `registryStatus` field to `ApiCollection` for tracking registry status.
- Introduced `RegistryBadge` component to display registry status in the UI.
- Updated `AuditData` and `ApiCollections` pages to show registry status using the new badge.
- Enhanced state management in `PersistStore` to include `collectionsRegistryStatusMap` for tracking registry statuses.
- Modified data fetching logic to populate registry status for collections.

### DIFF
--- a/apps/dashboard/src/main/java/com/akto/action/McpRegistryAction.java
+++ b/apps/dashboard/src/main/java/com/akto/action/McpRegistryAction.java
@@ -1,0 +1,187 @@
+package com.akto.action;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.bson.conversions.Bson;
+
+import com.akto.dao.ConfigsDao;
+import com.akto.dao.context.Context;
+import com.akto.dto.Config.ConfigType;
+import com.akto.dto.Config.McpRegistryConfig;
+import com.akto.dto.Config.McpRegistryConfig.McpRegistry;
+import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.Updates;
+
+public class McpRegistryAction extends UserAction {
+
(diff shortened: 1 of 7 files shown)


libs/utils/src/main/java/com/akto/mcp/McpRegistryUtils.java
```diff
@@ -0,0 +1,471 @@
...
+            return false;
+        }
+        
+        try {
+            // Normalize inputs
+            url = url.trim();
+            hostname = hostname.trim();
+            
+            // Remove protocol (http://, https://, etc.)
+            String urlWithoutProtocol = url.replaceFirst("^[a-zA-Z][a-zA-Z0-9+.-]*://", "");
```
Ark2307: use uri instead

libs/utils/src/main/java/com/akto/mcp/McpRegistryUtils.java
```diff
@@ -0,0 +1,471 @@
...
+            JsonNode remotesNode = server.get(JSON_FIELD_REMOTES);
+            if (remotesNode != null && remotesNode.isArray()) {
+                for (JsonNode remote : remotesNode) {
+                    if (remote == null) continue;
+                    
+                    JsonNode urlNode = remote.get(JSON_FIELD_URL);
+                    if (urlNode != null && !urlNode.isNull()) {
+                        String remoteUrl = urlNode.asText();
+                        if (remoteUrl != null && !remoteUrl.isEmpty()) {
+                            if (urlMatchesDomain(remoteUrl, hostname)) {
```
Ark2307: too much nested conditions, make one condition only for true

PR #3483, opened 2025-11-05
Add registry status management for API collections

krngrover94:
- Introduced `registryStatus` field in `DbAction` and `ApiCollection` to track MCP Registry status.
- Implemented `updateApiCollectionRegistryStatus` method in `ClientActor`, `DbActor`, and `DbLayer` for updating registry status.
- Added new action in `struts.xml` for updating API collection registry status.
- Integrated MCP Registry checks in `McpToolsSyncJobExecutor` to update API collections based on registry availability.
- Created `McpRegistryUtils` for handling MCP Registry queries and caching results.

### DIFF
--- a/apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/McpToolsSyncJobExecutor.java
+++ b/apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/McpToolsSyncJobExecutor.java
@@ -14,6 +14,7 @@ import com.akto.dto.traffic.SampleData;
 import com.akto.hybrid_parsers.HttpCallParser;
 import com.akto.log.LoggerMaker;
 import com.akto.log.LoggerMaker.LogDb;
+import com.akto.mcp.McpRegistryUtils;
 import com.akto.mcp.McpSchema;
 import com.akto.mcp.McpSchema.CallToolRequest;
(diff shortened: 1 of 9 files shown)


apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/McpToolsSyncJobExecutor.java
```diff
@@ -64,6 +65,13 @@ public class McpToolsSyncJobExecutor {
     public static final String LOCAL_IP = "127.0.0.1";
     private ServerCapabilities mcpServerCapabilities = null;
 
+    // Configuration for MCP Registry integration
+    private static final String ENV_ENABLE_REGISTRY_CHECK = "ENABLE_MCP_REGISTRY_CHECK";
+    private static final boolean ENABLE_REGISTRY_CHECK_DEFAULT = true;
+    private static final boolean enableRegistryCheck = Boolean.parseBoolean(
+        System.getenv().getOrDefault(ENV_ENABLE_REGISTRY_CHECK, String.valueOf(ENABLE_REGISTRY_CHECK_DEFAULT))
```
Ark2307, line 72: add try catch or normal simple check of value.equals(true) and use boolean so there is no chance of null value

PR #3489, opened 2025-11-06
prompt hardening ui and backend changes

shivam-rawat-akto:


apps/dashboard/src/main/java/com/akto/action/prompt_hardening/PromptHardeningAction.java
```diff
@@ -0,0 +1,455 @@
...
+        } catch (Exception e) {
+            e.printStackTrace();
+            addActionError(e.getMessage());
+            return ERROR.toUpperCase();
+        }
+    }
+
+    // Getters and Setters
+    public Map<String, Object> getPromptsObj() {
+        return promptsObj;
```
Ark2307: use lombok for getters and setters

libs/utils/src/main/java/com/akto/gpt/handlers/gpt_prompts/PromptHardeningTestHandler.java
```diff
@@ -0,0 +1,323 @@
...
+        result.put("agentResponse", rawResponse);
+        
+        return result;
+    }
+
+    /**
+     * Analyzes the agent response to determine if it's vulnerable
+     * This method applies detection rules to the response
+     */
+    public static BasicDBObject analyzeVulnerability(String agentResponse, BasicDBObject detectionRules) {
```
Ark2307, line 89: use llm only, something like MagicValidator

notshivansh (approved):

PR #4811, opened 2026-04-14
add support for collection filters in test results and issues

notshivansh:


TangoBeeAkto (dismissed):

apps/dashboard/src/main/java/com/akto/action/observe/InventoryAction.java
```diff
@@ -1303,6 +1307,10 @@ public String fetchTestedApisRanges(){
             } catch(Exception e){
             }
 
+            if (this.apiCollectionIds != null && !this.apiCollectionIds.isEmpty()) {
```
Ark2307: put it i guess, first filter out on collection and then apply sort, should be faster i guess

apps/dashboard/src/main/java/com/akto/action/testing/StartTestAction.java
```diff
@@ -1708,6 +1718,17 @@ public String fetchTestingRunsRanges(){
             pipeLine.add(
                 Aggregates.match(Filters.and(Filters.eq(TestingRunResultSummary.STATE, State.COMPLETED.toString()), Filters.gt(TestingRunResultSummary.START_TIMESTAMP, 0)))
             );
+            if (this.apiCollectionIds != null && !this.apiCollectionIds.isEmpty()) {
+                pipeLine.add(Aggregates.lookup(
+                        TestingRunDao.instance.getCollName(),
```
Ark2307, line 1723: can we add the same for context source too, so that agentic and api data dont overlap

apps/dashboard/src/main/java/com/akto/action/testing_issues/IssuesAction.java
```diff
@@ -489,6 +494,9 @@ public String fetchTestCoverageData() {
                 Filters.gte("time", startEpoch),
                 Filters.lte("time", endTimeStamp)
         );
+        if (filterCollectionsId != null && !filterCollectionsId.isEmpty()) {
```
Ark2307, line 497: create index here, no index is there for HistoricalData

apps/dashboard/src/main/java/com/akto/action/DashboardAction.java
```diff
@@ -160,13 +164,16 @@ public String fetchCriticalIssuesTrend(){
...
-                Filters.in(TestingRunIssues.TEST_RUN_ISSUES_STATUS, allowedStatus),
-                Filters.nin(TestingRunIssues.ID_API_COLLECTION_ID, demoCollections)
-        );
+        List<Bson> issueMatchParts = new ArrayList<>();
+        issueMatchParts.add(Filters.in(KEY_SEVERITY, severityToFetch));
+        issueMatchParts.add(Filters.gte(TestingRunIssues.CREATION_TIME, startTimeStamp));
+        issueMatchParts.add(Filters.lte(TestingRunIssues.CREATION_TIME, endTimeStamp));
+        issueMatchParts.add(Filters.in(TestingRunIssues.TEST_RUN_ISSUES_STATUS, allowedStatus));
+        issueMatchParts.add(Filters.nin(TestingRunIssues.ID_API_COLLECTION_ID, demoCollections));
+        if (apiCollectionIds != null && !apiCollectionIds.isEmpty()) {
```
Ark2307, line 173: sometimes you query on collectionIds, sometimes on "_id.apiinfokey.apiCollectionId" for the same collection, any specific reason?

libs/dao/src/main/java/com/akto/dao/testing_run_findings/TestingRunIssuesDao.java
```diff
@@ -191,16 +191,23 @@ public Map<ApiInfoKey, Map<String, Integer>> getSeveritiesMapForApiInfoKeys(Bson
...
-                Filters.lte(TestingRunIssues.LAST_SEEN, endTimeStamp),
-                Filters.gte(TestingRunIssues.LAST_SEEN, startTimeStamp),
-                Filters.nin("_id.apiInfoKey.apiCollectionId", deactivatedCollections)
-            )
-        ));
+        List<Bson> matchConditions = new ArrayList<>();
+        matchConditions.add(Filters.eq(TestingRunIssues.TEST_RUN_ISSUES_STATUS, "OPEN"));
+        matchConditions.add(Filters.lte(TestingRunIssues.LAST_SEEN, endTimeStamp));
+        matchConditions.add(Filters.gte(TestingRunIssues.LAST_SEEN, startTimeStamp));
+        matchConditions.add(Filters.nin("_id.apiInfoKey.apiCollectionId", deactivatedCollections));
```
Ark2307: use string literals

apps/dashboard/src/main/java/com/akto/action/observe/InventoryAction.java
notshivansh: Uploading Screen Recording 2026-04-14 at 4.14.25 PM.mov…

nope, doesn't matter.

notshivansh: https://github.com/user-attachments/assets/42eff098-b2ee-4f3f-a1aa-1baff5f8efec

apps/dashboard/src/main/java/com/akto/action/testing/StartTestAction.java
notshivansh: the filter is on `apiCollectionIds` field, which is being sent by the user.
If I understand correctly, context source would help, if we were calculating the collection ids, right ?
we can add it as an additional check, but would be redundant, because we're already showing the filtered collections in UI, where this field is coming from.

libs/dao/src/main/java/com/akto/dao/testing_run_findings/TestingRunIssuesDao.java
notshivansh: done

apps/dashboard/src/main/java/com/akto/action/DashboardAction.java
notshivansh: fixed, all use `_id.apiinfokey.apiCollectionId `

apps/dashboard/src/main/java/com/akto/action/testing_issues/IssuesAction.java
notshivansh: added

apps/dashboard/src/main/java/com/akto/action/observe/InventoryAction.java
notshivansh: The documents scanned and the indices scanned remain the same in both the iterations, the time is different, might be an internal caching thing for the second query.

Have updated the pipeline, anyways.

Ark2307 (approved):

PR #4863, opened 2026-04-17
refactor GraphMetric to remove backgroundColor prop and set default b…

harshithb3304:
…ackground color; update grid line width and add exporting options

### DIFF
--- a/apps/dashboard/web/polaris_web/web/src/apps/dashboard/components/GraphMetric.jsx
+++ b/apps/dashboard/web/polaris_web/web/src/apps/dashboard/components/GraphMetric.jsx
@@ -10,7 +10,7 @@ require("highcharts/modules/mouse-wheel-zoom")(Highcharts);
 
 function GraphMetric(props) {
 
-    const { height, backgroundColor, data, inputMetrics, title, text, defaultChartOptions, subtitle, timezoneOffsetMinutes } = props;
+    const { height, data, inputMetrics, title, text, defaultChartOptions, subtitle, timezoneOffsetMinutes } = props;
     const chartComponentRef = useRef(null)
 
     const fillColor = {
@@ -59,7 +59,7 @@ function GraphMetric(props) {
         chart: {
             type: 'spline',
             height: `${height}px`,
-            backgroundColor,
+            backgroundColor: '#ffffff',
             animation: false,
             spacingTop: 20,
             spacingLeft: 16,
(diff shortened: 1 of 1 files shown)


apps/dashboard/web/polaris_web/web/src/apps/dashboard/components/GraphMetric.jsx
```diff
@@ -59,7 +59,7 @@ function GraphMetric(props) {
         chart: {
             type: 'spline',
             height: `${height}px`,
-            backgroundColor,
+            backgroundColor: '#ffffff',
```
utafrali, line 62: Hardcoding `backgroundColor: '#ffffff'` silently breaks any caller that wants a custom background. `MetricChart.jsx` passes `background-color="#000000"` (already a separate bug with the hyphenated prop name, but still an indicator that dark backgrounds were intended). Since `...defaultChartOptions` is spread at the top level, a caller cannot easily override a nested `chart.backgroundColor` without replacing the entire `chart` config block.

If the goal is simply to standardize the default, keep the prop with a default value instead:
```jsx
const { height, backgroundColor = '#ffffff', data, ... } = props;
```
This preserves backward compatibility while establishing the white default.

apps/dashboard/web/polaris_web/web/src/apps/dashboard/components/GraphMetric.jsx
```diff
@@ -125,6 +125,11 @@ function GraphMetric(props) {
                 opposite: true,
             })),
         ],
+        exporting: {
```
utafrali, line 128: The `exporting` block is placed **before** `...defaultChartOptions`. JavaScript's object spread is a shallow merge, so if any caller passes `exporting` inside their `defaultChartOptions`, the entire `exporting` object (including the `backgroundColor` fix) will be completely replaced with no warning.

If the intent is to always export with a white background regardless of caller customization, move the block after the spread:
```js
...defaultChartOptions,
exporting: {
    ...(defaultChartOptions?.exporting ?? {}),
    chartOptions: {
        chart: { backgroundColor: '#ffffff' }
    }
},
```
This both preserves caller `exporting` settings and guarantees the background color is applied.

utafrali (commented):
The refactor works for current callers (all of which happen to use `'#ffffff'`), but hardcoding the background color removes legitimate flexibility and creates a silent trap for future callers. The `exporting` config placement before `...defaultChartOptions` also means it can be silently wiped by any caller that passes their own `exporting` key.

PR #4887, opened 2026-04-21
Supporting in service for validating request

Ark2307:


rakshak-akto (approved):

apps/guardrails-service/container/src/pkg/validator/service.go
```diff
@@ -164,7 +164,7 @@ func filterPoliciesByMcpServer(policies []types.Policy, mcpServerName string) []
 	mcpServerNameLower := strings.ToLower(mcpServerName)
 	filtered := make([]types.Policy, 0, len(policies))
 	for _, policy := range policies {
-		if policy.IsYamlPolicy {
+		if policy.IsYamlPolicy || policy.ApplyToAllServers {
```
rakshak-akto, line 167: make changed in go.mod and go.sum to use updated version of mcp-endpoint-shield go lib dependency

rakshak-akto (changes requested):

rakshak-akto (approved):

PR #4909, opened 2026-04-22
Add sequences

gauravakto:


apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -699,6 +702,38 @@ public void run() {
...
     }
 
+    public String writeApiSequences() {
+        if (apiSequencesList == null || apiSequencesList.isEmpty()) {
+            return Action.SUCCESS.toUpperCase();
+        }
+        try {
+            List<WriteModel<ApiSequences>> writeModels = new ArrayList<>();
+            for (ApiSequences seq : apiSequencesList) {
+                float probability = seq.getPrevStateCount() > 0
```
ayushaga14: take summation and calc

libs/dao/src/main/java/com/akto/dto/ApiSequences.java
```diff
@@ -0,0 +1,80 @@
...
+
+    // Example: No of times transition from a->b was seen
+    int transitionCount;
+    public static final String TRANSITION_COUNT = "transitionCount";
+
+    // Example: No of times the path "a" was seen
+    int prevStateCount;
+    public static final String PREV_STATE_COUNT = "prevStateCount";
+
+    float precedenceScore;
```
ayushaga14, line 39: is this field being used?

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -699,6 +703,50 @@ public void run() {
         return Action.SUCCESS.toUpperCase();
     }
 
+    public String writeApiSequences() {
+        if (apiSequencesList == null || apiSequencesList.isEmpty()) {
+            return Action.SUCCESS.toUpperCase();
+        }
+        try {
+            List<WriteModel<ApiSequences>> writeModels = new ArrayList<>();
+            for (ApiSequences seq : apiSequencesList) {
```
ayushaga14, line 712: what can be the max size of apiSequencesList

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -699,6 +703,50 @@ public void run() {
...
+                Bson filter = Filters.and(
+                        Filters.eq(ApiSequences.API_COLLECTION_ID, seq.getApiCollectionId()),
+                        Filters.eq(ApiSequences.PATHS, seq.getPaths())
```
ayushaga14, line 715: is paths array something like [a,b], eq filter can maybe not work sometimes if order is different in db [b,a]

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
gauravakto: mini runtime sending in batches of 100 maximum

libs/dao/src/main/java/com/akto/dto/ApiSequences.java
gauravakto: yes this will be used to determine how confident we are about this sequence being correct.

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
gauravakto: order will be maintained.

ayushaga14 (approved):

PR #4928, opened 2026-04-23
Rate limit optimisations scram

ayushaga14:
## Summary

Optimises rate-limit detection in `threat-detection` by moving **all Redis operations off the Kafka hot path** — replacing synchronous `EVALSHA` calls (which halved throughput from ~5500 to ~2500 msg/sec) with a fully async Redis Stream pipeline.

### Key Changes

- **Zero sync Redis calls on hot path**: Kafka consumer now only does an async `XADD` to a Redis Stream per message
- **Batch Lua processing**: A background stream consumer batches up to 500 messages and processes them in a single `EVALSHA` call — amortising network round trips
- **CMS moved to Redis (RedisBloom)**: Replaced in-memory `CountMinSketch` with `CMS.INITBYPROB` / `CMS.INCRBY` / `CMS.QUERY` in Redis — centralised, no cross-pod sync needed
- **Distribution race condition fixed**: CMS increment + distribution bucket update now happen atomically in the same Lua script — eliminates negative bucket counts
- **Mitigation check in Lua**: `isActorInMitigationPeriod` moved into the Lua script to avoid extra sync Redis calls
- **Bucket boundary fix**: Changed `count < max` to `count <= max` in Lua to match inclusive bucket ranges from `ThreatApiDistributionUtils`

## Architecture Flow

```
┌─────────────────────────────────────────────────────────────────┐
│                        KAFKA CONSUMER                           │
│                   (MaliciousTrafficDetectorTask)                │
│                                                                 │
│  For each message:                                              │
│    1. Parse HttpResponseParams                                  │
│    2. Run filter matching (malicious traffic detection)         │
│    3. If apiDistributionEnabled:                                │
│       └─ async XADD to "threat_input_stream"                   │
│          (ipApiCmsKey, apiKey, epochMin, threshold,             │
│           host, accountId, rateLimitWindow)                     │
│                                                                 │
│  ⚡ Zero sync Redis calls = full ~5500 msg/sec throughput       │
└──────────────────────────┬──────────────────────────────────────┘
                           │ async XADD
                           ▼
┌─────────────────────────────────────────────────────────────────┐
│                    REDIS STREAM                                  │
│               "threat_input_stream"                              │
│                                                                 │
│  Messages accumulate here. Consumer group "threat_group"         │
│  ensures each message is processed exactly once.                │
└──────────────────────────┬──────────────────────────────────────┘
                           │ XREADGROUP (batch up to 500)
                           ▼
┌─────────────────────────────────────────────────────────────────┐
│                STREAM CONSUMER (Background Thread)              │
│                  (DistributionStreamConsumer)                    │
│                                                                 │
│  1. XREADGROUP COUNT 500 from "threat_input_stream"             │
│  2. Group messages into batch                                   │
│  3. Single EVALSHA call with entire batch                       │
│     └─ 500 messages × 1 round trip = 500x fewer round trips    │
└──────────────────────────┬──────────────────────────────────────┘
                           │ EVALSHA
                           ▼
┌─────────────────────────────────────────────────────────────────┐
│              LUA SCRIPT (async_threat_process.lua)               │
│              Runs atomically in Redis (single-threaded)          │
│                                                                 │
│  For each message in batch:                                     │
│    1. CMS.INITBYPROB (if key doesn't exist)                    │
│    2. CMS.INCRBY → increment count for IP+API in current min   │
│    3. EXPIRE on CMS key (8hr TTL)                               │
│    4. CMS.QUERY across sliding window → slidingCount            │
│    5. Distribution update:                                      │
│       ├─ getBucketLabel(count) → newBucket                      │
│       ├─ getBucketLabel(count-1) → oldBucket                    │
│       ├─ If count == 1: HINCRBY dist newBucket +1               │
│       └─ If oldBucket ≠ newBucket: HINCRBY old -1, new +1      │
│    6. SADD to active APIs set                                   │
│    7. Threshold check:                                          │
│       ├─ EXISTS mitigation key → skip if in cooldown            │
│       ├─ If slidingCount > threshold:                           │
│       │   └─ SETEX mitigation key (5 min TTL)                  │
│       │   └─ Add to breaches list                               │
│       └─ Return breaches to Java                                │
│                                                                 │
│  ✅ CMS + Distribution atomic = no race condition               │
│  ✅ No negative bucket counts                                   │
└──────────────────────────┬──────────────────────────────────────┘
                           │ breaches returned
                           ▼
┌─────────────────────────────────────────────────────────────────┐
│            BREACH EVENT HANDLER (in StreamConsumer)              │
│                                                                 │
│  For each breach:                                               │
│    1. Parse ipApiCmsKey → IP, apiCollectionId, url, method      │
│    2. Build MaliciousEventMessage (from stream message fields)  │
│    3. Push to internal Kafka topic                              │
└─────────────────────────────────────────────────────────────────┘
```

### Redis Data Model

```
cms|{epochMin}                          → RedisBloom CMS sketch (8hr TTL)
dist|{windowSize}|{windowStart}|{api}   → Hash {b1: count, b2: count, ...}
distApis|{windowSize}|{windowStart}     → Set of active API keys
mitigate|{ipApiCmsKey}                  → String "1" (5 min TTL)
threat_input_stream                     → Redis Stream (consumer group: threat_group)
```

### Multi-Pod Behaviour
- All pods share the same Redis instance
- Kafka is IP-partitioned → same IP always routes to same pod
- Each pod runs 1 stream consumer with a unique consumer ID (instanceId)
- Consumer group ensures each stream message is processed exactly once
- CMS operations are commutative (order-independent) — safe across pods

### Files Changed
- **`async_threat_process.lua`** — New mega Lua script (CMS + distribution + mitigation + threshold)
- **`cms_increment_and_query.lua`** / **`cms_range_query.lua`** — Retained for `ParamEnumerationDetector`
- **`DistributionStreamConsumer.java`** — New stream consumer with batch EVALSHA
- **`DistributionCalculator.java`** — Simplified to async XADD only
- **`CmsCounterLayer.java`** — Rewritten to use Redis-backed CMS via Lua scripts
- **`MaliciousTrafficDetectorTask.java`** — Hot path no longer calls sync Redis
- **`CmsDataForwardLayer.java`** / **`distribution_update.lua`** — Deleted (obsolete)

## Test plan
- [x] CMS verification benchmark: counts match expected values across IPs and APIs
- [x] Distribution verification: correct bucket assignments (b1=1, b2=1, b3=1, b5=1 for known inputs), zero negative counts
- [x] Stream consumer: all messages consumed (lag=0, pending=0)
- [x] Multi-API distribution: correct per-API distributions across all window sizes (5, 15, 30 min)
- [ ] Throughput benchmark: verify ~5500 msg/sec with aggregationRulesEnabled=true
- [ ] Multi-pod deployment test


gauravakto (commented):
## Code Review

### 🔴 Critical

**[`async_threat_process.lua`] Lua script will block the Redis event loop**

In a batch of 500 messages, worst case this script executes up to **500 × (5 + 15 + 30) = 25,000 `CMS.QUERY` calls** (across all 3 tumbling windows per message), plus 500 `CMS.INCRBY` calls, plus HSET/HINCRBY for distribution. Redis is single-threaded — one Lua script execution blocks **all other Redis commands** for its entire duration. At 5500 msg/sec, this is a 90ms–500ms stall risk per Lua invocation, which defeats the goal of keeping the hot path fast.

Consider: breaking the batch into smaller Lua calls, pre-aggregating per-minute CMS counts in Java and passing pre-summed counts into the script, or processing distribution updates separately outside the hot Lua path.

---

**[`DistributionStreamConsumer.java`] No pending message recovery on restart**

`run()` always reads with `lastConsumed()` (`>` in XREADGROUP), which only delivers **new** messages. If the pod crashes or `processBatch()` throws after `XREADGROUP` but before `XACK`, those messages enter the **Pending Entries List (PEL)** and are **never retried** — there is no `XAUTOCLAIM` or `XPENDING`-based recovery on startup.

Fix: on startup, first drain pending messages using `XReadArgs.StreamOffset.from(STREAM, "0-0")` until an empty batch is returned, then switch to `lastConsumed()`.

---

### 🟠 Important

**[`async_threat_process.lua`] Redis Cluster incompatibility**

All CMS keys (`"cms|" .. currentMin`) are constructed **inside** the Lua script and never declared in `KEYS[]`. Redis Cluster requires all keys a script touches to be pre-declared in `KEYS[]` — otherwise it cannot route the command correctly. If the deployment ever uses Redis Cluster, every `EVALSHA` call will fail with a `CROSSSLOT` error.

Even in standalone mode, this violates the Redis Lua scripting contract. Consider passing the CMS keys via `KEYS[]` or explicitly documenting that Cluster mode is unsupported.

---

**[`DistributionStreamConsumer.java`] Stream trim can drop unprocessed messages under lag**

At 5500 msg/sec, `MAX_STREAM_LENGTH = 500_000` gives roughly ~90 seconds of buffer. During a Redis slowdown, a slow Lua execution, or a consumer restart, the stream can easily exceed this limit. `XTRIM MAXLEN` removes old entries **regardless of PEL status** — unprocessed messages can be permanently deleted before they are consumed.

Mitigations: (a) increase `MAX_STREAM_LENGTH` significantly, (b) switch to `XTRIM MAXLEN ~` (approximate), or (c) only trim based on the last-ACKed message ID rather than a fixed length cap.

---

**[`Main.java`] ExecutorService for stream consumer is never shut down**

`startDistributionStreamConsumers()` creates a `newFixedThreadPool(1)` but the reference is local and discarded. There is no shutdown hook — on JVM shutdown the consumer thread will be killed mid-batch, potentially leaving messages stuck in PEL.

Store the executor reference and register a `Runtime.getRuntime().addShutdownHook(...)` that calls `executor.shutdownNow()` to let the consumer thread exit cleanly.

---

**[General] No tests for new code**

`CmsCounterLayerTest` (221 lines) and `DistributionCalculatorTest` (229 lines) are deleted. The new `DistributionStreamConsumer`, `async_threat_process.lua` (bucket transitions, mitigation gating, sliding window math), and the Redis-backed `CmsCounterLayer` have **zero test coverage**. The test plan marks stream consumer and multi-pod tests as pending.

This is a significant risk for a performance-critical path change. At minimum, the Lua script logic — bucket boundary transitions, mitigation key lifecycle, breach detection — should have integration tests before merge.

---

### 🟡 Minor

**[`DistributionDataForwardLayer.java`] `sentWindows` grows unbounded**

The in-memory `Set<String> sentWindows` accumulates window keys indefinitely. Over 8 hours across 3 window sizes and many API keys, this set can grow large. On pod restart it resets (acceptable per the comment), but in a long-running pod old entries are never evicted.

Consider capping the set size or evicting entries older than `LOOKBACK_HOURS`.

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/CmsCounterLayer.java
```diff
@@ -1,188 +1,144 @@
...
     private static CmsCounterLayer DEFAULT_INSTANCE;
 
-    /**
-     * Constructor for creating new instances with custom cache and Redis key prefix.
-     */
-    public CmsCounterLayer(CounterCache cache, String redisKeyPrefix) {
-        this.cache = cache;
-        this.redisKeyPrefix = redisKeyPrefix;
+    public CmsCounterLayer(RedisClient redisClient, String keyPrefix) {
+        this.connection = redisClient.connect();
```
gauravakto, line 40: null check or try/catch ?

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/DistributionCalculator.java
```diff
@@ -1,148 +1,65 @@
...
-     *   check here     check here    check here
-     */
-    private long getCountInRange(String ipApiCmsKey, long startEpochMin, long endEpochMin) {
-        long sum = 0;
-        for (long ts = startEpochMin; ts <= endEpochMin; ts++) {
-            sum += cmsCounterLayer.estimateCount(ipApiCmsKey, String.valueOf(ts));
-        }
-        return sum;
+    public DistributionCalculator(RedisClient redisClient) {
+        this.redis = redisClient.connect();
```
gauravakto, line 27: wrap in try/catch?

apps/threat-detection/src/main/java/com/akto/threat/detection/Main.java
```diff
@@ -111,13 +112,26 @@ public static void main(String[] args) throws Exception {
...
+    DistributionDataForwardLayer distributionDataForwardLayer = localRedis != null
+        ? new DistributionDataForwardLayer(localRedis)
+        : null;
 
     boolean apiDistributionEnabled = Utils.apiDistributionEnabled(localRedis != null, System.getenv().getOrDefault("API_DISTRIBUTION_ENABLED", "true").equals("true"));
 
     triggerDistributionDataForwardCron(apiDistributionEnabled, distributionDataForwardLayer);
 
+    com.akto.threat.detection.kafka.KafkaProtoProducer internalKafkaProducer =
+        new com.akto.threat.detection.kafka.KafkaProtoProducer(internalKafka);
```
gauravakto: sendTomalicoiusBackend must be using a producer as well ? see how that is made, this one doesn't seem to be configuring anything.

apps/threat-detection/src/main/java/com/akto/threat/detection/Main.java
```diff
@@ -158,12 +172,20 @@ public static void triggerDistributionDataForwardCron(boolean apiDistributionEna
...
     try {
-        //distributionDataForwardLayer.sendLastFiveMinuteDistributionData();
+        distributionDataForwardLayer.sendLastFiveMinuteDistributionData();
     } catch (Exception e) {
         logger.errorAndAddToDb(e,"Error scheduling relayApiCountInfoCron : {} ");
     }
   }
 
+  public static void startDistributionStreamConsumers(RedisClient redisClient,
+      com.akto.threat.detection.kafka.KafkaProtoProducer internalKafkaProducer, String instanceId) {
```
gauravakto: move imports to top.

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/DistributionStreamConsumer.java
```diff
@@ -0,0 +1,283 @@
...
+import io.lettuce.core.api.StatefulRedisConnection;
+import io.lettuce.core.api.sync.RedisCommands;
+
+public class DistributionStreamConsumer implements Runnable {
+
+    private static final LoggerMaker logger = new LoggerMaker(DistributionStreamConsumer.class, LogDb.THREAT_DETECTION);
+
+    private static final String GROUP_NAME = "threat_group";
+    private static final int BATCH_SIZE = 500;
+    private static final int CMS_TTL_SECONDS = 8 * 60 * 60;
```
gauravakto: time units pe can comments 8 hours..

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/DistributionStreamConsumer.java
```diff
@@ -0,0 +1,283 @@
...
+                    .setLatestApiEndpoint(url)
+                    .setCategory(IP_API_RATE_LIMIT_FILTER.getInfo().getCategory().getName())
+                    .setSubCategory(IP_API_RATE_LIMIT_FILTER.getInfo().getSubCategory())
+                    .setSeverity(IP_API_RATE_LIMIT_FILTER.getInfo().getSeverity())
+                    .setMetadata(maliciousReq.getMetadata())
+                    .setType("Rule-Based")
+                    .setSuccessfulExploit(false)
+                    .setStatus(status)
+                    .setHost(host)
+                    .setContextSource(com.akto.util.enums.GlobalEnums.CONTEXT_SOURCE.API.name())
```
gauravakto: imports to top everywhere

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/DistributionStreamConsumer.java
```diff
@@ -0,0 +1,283 @@
...
+                    .setLatestApiCollectionId(apiCollectionId)
+                    .setLatestApiIp(actor)
+                    .setLatestApiPayload("")
+                    .setLatestApiMethod(method)
+                    .setLatestApiEndpoint(url)
+                    .setCategory(IP_API_RATE_LIMIT_FILTER.getInfo().getCategory().getName())
+                    .setSubCategory(IP_API_RATE_LIMIT_FILTER.getInfo().getSubCategory())
+                    .setSeverity(IP_API_RATE_LIMIT_FILTER.getInfo().getSeverity())
+                    .setMetadata(maliciousReq.getMetadata())
+                    .setType("Rule-Based")
```
gauravakto: dynamic ratelimting should be anomally based ?

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/DistributionStreamConsumer.java
```diff
@@ -0,0 +1,283 @@
...
+            internalKafka.send(KafkaTopic.ThreatDetection.ALERTS, envelope);
+            logger.debugAndAddToDb("Breach event pushed for actor: " + actor);
+
+        } catch (Exception e) {
+            logger.errorAndAddToDb(e, "Error handling breach: " + breachData);
+        }
+    }
+
+    private void trimStreamIfNeeded() {
+        long now = System.currentTimeMillis();
```
gauravakto, line 289: ask claude to add comment why this is needed ?

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/DistributionStreamConsumer.java
```diff
@@ -0,0 +1,283 @@
...
+                logger.errorAndAddToDb("Invalid ipApiCmsKey format: " + ipApiCmsKey);
+                return;
+            }
+            int apiCollectionId = Integer.parseInt(parts[1]);
+            String url = parts[3];
+            String method = parts[4];
+
+            String status = com.akto.util.ThreatDetectionConstants.ACTIVE;
+
+            Metadata metadata = Metadata.newBuilder()
```
gauravakto, line 232: re-use  function? utils.buildMaliciosuRequest?

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/ParamEnumerationDetector.java
```diff
@@ -33,126 +30,81 @@ public class ParamEnumerationDetector {
...
     }
 
-    /**
-     * Initialize the singleton instance with custom window size.
-     *
-     * @param redisClient      Redis client
-     * @param threshold        Number of unique values to trigger detection
-     * @param windowSizeMinutes Size of the sliding window in minutes
-     */
     public static void initialize(RedisClient redisClient, int threshold, int windowSizeMinutes) {
```
gauravakto, line 48: is this INSTANCE=null then this feature should not run at all right ?

gauravakto (approved):

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/CmsCounterLayer.java
ayushaga14: already null check in caller

apps/threat-detection/src/main/java/com/akto/threat/detection/ip_api_counter/DistributionCalculator.java
ayushaga14: already there in caller

apps/threat-detection/src/main/java/com/akto/threat/detection/Main.java
ayushaga14: this is being used by redis consumer to send rate limit breach messages in kafka

PR #4933, opened 2026-04-23
fix prettified endpoint

nayanakto:


apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/observe/GetPrettifyEndpoint.jsx
```diff
@@ -7,11 +7,11 @@ import { isAgenticSecurityCategory, isMCPSecurityCategory, isEndpointSecurityCat
 
 export const getMethod = (url, method) => {
     if(isMCPSecurityCategory() || isAgenticSecurityCategory() || isEndpointSecurityCategory()){
-        if(url.includes("tool")){
+        if(/\/tools?\//i.test(url)){
```
TangoBeeAkto: What about just using path.split("/").includes("tool")?

TangoBeeAkto: i guess `(/\/tools?\//i.test(url))` won't handle this case: "/tool"

nayanakto: the regex ends with ?/ which means "s" is optional. anyways, mcp servers uses "tools" and not "tool"

TangoBeeAkto (approved):

PR #4938, opened 2026-04-24
add before/after tool call hooks

DharunThota:


apps/mcp-endpoint-shield/gemini-cli-hooks/akto-validate-post-tool-wrapper.ps1
```diff
@@ -0,0 +1,17 @@
+# Auto-generated wrapper for Akto guardrails hook
+# Generated by MCP Endpoint Shield installer
+
+$env:MODE = "atlas"
+$env:AKTO_DATA_INGESTION_URL = "{{AKTO_DATA_INGESTION_URL}}"
+$env:AKTO_SYNC_MODE = "true"
+$env:AKTO_TIMEOUT = "5"
+$env:AKTO_CONNECTOR = "gemini_cli"
+$env:AKTO_CONNECTOR_VALUE = "geminicli"
```
rakshak-akto, line 9: what's the difference between these two? and why different values?

rakshak-akto (approved):

apps/mcp-endpoint-shield/gemini-cli-hooks/akto-validate-post-tool-wrapper.ps1
DharunThota: AKTO_CONNECTOR is used in the request as a query param, whereas AKTO_CONNECTOR_VALUE is used in the tags and to form the hostname for atlas mode. I used the Claude CLI Hooks as reference, and it followed the same naming convention

PR #4944, opened 2026-04-27
Feature/collection skills support

rahul-akto:


apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -617,6 +613,16 @@ public String fetchApiRateLimits() {
...
+    public String fetchApiInfosByCollection() {
+        try {
+            apiInfos = DbLayer.fetchApiInfosByCollection(apiCollectionId);
+        } catch (Exception e) {
+            loggerMaker.errorAndAddToDb(e, "error in fetchApiInfosByCollection " + e.toString());
+            return Action.ERROR.toUpperCase();
+        }
+        return Action.SUCCESS.toUpperCase();
+    }
+
```
notshivansh, line 625: see if already an API exists for this, seems like a use case we might already be serving.

libs/utils/src/main/java/com/akto/data_actor/DbLayer.java
```diff
@@ -413,6 +411,13 @@ public static List<ApiInfo> fetchApiInfos() {
         return ApiInfoDao.instance.findAll(new BasicDBObject(), Projections.exclude(ApiInfo.RATELIMITS));
     }
 
+    public static List<ApiInfo> fetchApiInfosByCollection(int apiCollectionId) {
+        return ApiInfoDao.instance.findAll(
+            Filters.eq("_id.apiCollectionId", apiCollectionId),
+            Projections.exclude(ApiInfo.RATELIMITS)
+        );
+    }
+
```
notshivansh, line 420: need limit in query.

.gitignore
```diff
@@ -1,3 +1,4 @@
+start.sh
```
notshivansh, line 1: why this ?

libs/utils/src/main/java/com/akto/data_actor/DbLayer.java
rahul-akto: there will be limited api info for single collection and i need to fetch all api_info fir given collection so that have create

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
rahul-akto: there is no any api i guess which serve this

.gitignore
rahul-akto: have script file to start application and want to add in gitignore

notshivansh (approved):

PR #4978, opened 2026-04-29
run both threat algos

ayushaga14:


gauravakto (approved):

gauravakto:
test with a custom policy as well both kinds normal policy and aggregation based

PR #4985, opened 2026-04-29
feat: block personal accounts via policy name in guardrails service

rahul-akto:
- Detect block_personal_account guardrail by policy name instead of a boolean field, using the already-parsed policies slice (removes rawGuardrailsPolicy struct and second JSON unmarshal)
- Add browser-llm-account-type=personal as a blocking signal alongside login-user-email-type
- Move personal account check after getCachedPolicies so the flag is populated on the first request
- Add FetchApiCollections to db abstractor client for collection tag cache
- Rename CollectionTagRefreshIntervalMin -> CollectionRefreshIntervalMin (env: COLLECTION_REFRESH_INTERVAL_MIN)

### DIFF
--- a/apps/guardrails-service/container/src/pkg/validator/service.go
+++ b/apps/guardrails-service/container/src/pkg/validator/service.go
@@ -415,6 +415,58 @@ func (s *Service) getLoginUserEmailType(params *models.ValidateRequestParams) st
 	return v
 }
 
+// TODO: move reportAndBlockPersonalAccount to mcp library so threat reporting
+// and validation live in one place alongside other policy enforcement.
+func (s *Service) reportAndBlockPersonalAccount(_ context.Context, params *models.ValidateRequestParams, payloadToValidate, sessionID, requestID string) *mcp.ValidationResult {
+	blockReason := "Blocked: personal accounts are not permitted by guardrail policy"
+
+	if s.sessionMgr != nil && sessionID != "" {
+		s.sessionMgr.TrackResponse(sessionID, requestID, blockReason, true)
+		s.sessionMgr.UpdateBlockedReason(sessionID, blockReason)
+	}
+
+	if !params.EffectiveSkipThreat() {
+		reqHeaders := make(map[string]string)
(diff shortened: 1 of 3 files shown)


apps/guardrails-service/container/src/pkg/validator/service.go
```diff
@@ -296,6 +311,162 @@ func (s *Service) refreshPolicies() ([]types.Policy, map[string]*types.AuditPoli
 	return policies, auditPolicies, compiledRules, hasAuditRules, nil
 }
 
+func hasBlockPersonalAccountPolicy(policies []types.Policy) bool {
+	for _, p := range policies {
+		if p.Info.Name == "block_personal_account" {
```
nayanakto, line 316: how are we adding this from the guardrails policy?

rakshak-akto (approved):

apps/guardrails-service/container/src/pkg/validator/service.go
```diff
@@ -296,6 +311,162 @@ func (s *Service) refreshPolicies() ([]types.Policy, map[string]*types.AuditPoli
...
+		reqHeaders := make(map[string]string)
+		if params.RequestHeaders != "" {
+			json.Unmarshal([]byte(params.RequestHeaders), &reqHeaders)
+		}
+		statusCode := 0
+		if params.StatusCode != "" {
+			fmt.Sscanf(params.StatusCode, "%d", &statusCode)
+		}
+		go func() {
+			if err := mcp.ReportThreat(
```
nayanakto, line 438: what is the filterId we are passing?
without that threats wont be visible.

rahul-akto: i have tested it's visible on dashboard .

apps/guardrails-service/container/src/pkg/validator/service.go
rahul-akto: since we don;t have support for apply guardrails policy that block user on the basis of account type logged in so that's why have done like this

PR #4993, opened 2026-04-30
refactor: clean up MaliciousTrafficDetectorTask and Kafka config layer

gauravakto:
## Summary

This PR refactors the `threat-detection` module focusing on `MaliciousTrafficDetectorTask` and the Kafka configuration layer. No behavioural changes — pure structural cleanup.

### Kafka config (`libs/utils`)
- Added `fetchMaxBytes` field to `KafkaConsumerConfig` so consumers can declare their fetch size in config rather than hardcoding it in `Properties` setup
- Added `toConsumerProperties()` to `KafkaConfig` — single place that translates the typed config object into the `Properties` map the Kafka client requires
- Fixed a pre-existing bug in `AbstractKafkaConsumerTask`: key deserializer was incorrectly set to the value serializer

### `AbstractKafkaConsumerTask`
- Consumer construction now uses `kafkaConfig.toConsumerProperties()` — no more duplicate `Properties` setup in subclasses
- Added `beforePollLoop()` hook for subclasses to run setup before polling starts
- Removed `if (records.isEmpty()) continue` guard — was preventing metrics logging during low-traffic periods in `MaliciousTrafficDetectorTask`

### `MaliciousTrafficDetectorTask`
- Now extends `AbstractKafkaConsumerTask<byte[]>` — eliminates duplicate Kafka consumer setup and the polling loop boilerplate
- Removed `kafkaPollingEnabled` static boolean (read once at startup, always true, not a real kill switch)
- Extracted `FilterCache` class (`cache/`) — owns filter state (`apiFilters`, `successfulExploitFilters`, `ignoredEventFilters`), TTL-based refresh, and OpenAPI schema Redis cache. Fixes a bug where `successfulExploitFilters` and `ignoredEventFilters` were `static` mutable lists
- Moved `handleSchemaConformFilter` to `RequestValidator` where the rest of schema validation logic lives
- Moved `getRedactionType` + `extractHostFromHeaders` to `Utils` — stateless helpers reusable by other components (e.g. `HyperscanEventHandler`)
- Removed unused `trackAndLogValidHostname` + `isValidHostname` methods and their associated dead state fields
- Extracted per-minute metrics logging into `logRecordsPerMin(int count)`
- Topic name `"akto.api.logs2"` moved to `KafkaTopic.TRAFFIC_LOGS` constant
- Removed all inline `java.util.*` imports, standardised at top

### Other
- Removed `CLAUDE.md` from `.gitignore` — the file contains module architecture docs useful to all contributors
- `Main.java` traffic Kafka config now explicitly sets `fetchMaxBytes = 100MB` (was hardcoded inside `MaliciousTrafficDetectorTask` constructor previously)

## Review notes
- `buildHttpResponseParam` intentionally stays in `MaliciousTrafficDetectorTask` — it reuses static `requestParams`/`responseParams` objects for performance on the hot Kafka path (single-threaded by design, documented in CLAUDE.md)
- `FilterCache.getApiSchema` handles `null` Redis connection gracefully — falls back to direct DB fetch when Redis is unavailable
- Schema conformance detection (`SchemaConform` category) remains gated to specific account IDs and is unchanged

## Test plan
- [ ] Verify Kafka consumer starts and subscribes to `akto.api.logs2` correctly
- [ ] Verify filters are fetched and bucketed correctly on first poll and after TTL expiry
- [ ] Verify `fetchMaxBytes` is applied for the traffic consumer (check consumer config logs at startup)
- [ ] Verify threat detection events are still published correctly for both YAML filter and Hyperscan paths
- [ ] Verify redaction logic behaves correctly for all three `RedactionType` values

### DIFF
--- a/apps/threat-detection/src/main/java/com/akto/threat/detection/tasks/AbstractKafkaConsumerTask.java
+++ b/apps/threat-detection/src/main/java/com/akto/threat/detection/tasks/AbstractKafkaConsumerTask.java
@@ -1,6 +1,8 @@
 package com.akto.threat.detection.tasks;
 
 import com.akto.kafka.KafkaConfig;
+import com.akto.log.LoggerMaker;
+import com.akto.log.LoggerMaker.LogDb;
 
 import java.time.Duration;
 import java.util.Collections;
@@ -12,13 +14,20 @@ import org.apache.kafka.clients.consumer.KafkaConsumer;
 
 public abstract class AbstractKafkaConsumerTask<V> implements Task {
 
+  private static final LoggerMaker logger = new LoggerMaker(AbstractKafkaConsumerTask.class, LogDb.THREAT_DETECTION);
+
   protected Consumer<String, V> kafkaConsumer;
   protected KafkaConfig kafkaConfig;
   protected String kafkaTopic;
+  protected String instanceId;
 
-  public AbstractKafkaConsumerTask(KafkaConfig kafkaConfig, String kafkaTopic) {
+  private int recordsReadCount = 0;
+  private long lastRecordCountLogTime = System.currentTimeMillis();
+
+  public AbstractKafkaConsumerTask(KafkaConfig kafkaConfig, String kafkaTopic, String instanceId) {
     this.kafkaTopic = kafkaTopic;
     this.kafkaConfig = kafkaConfig;
+    this.instanceId = instanceId;
     this.kafkaConsumer = new KafkaConsumer<>(kafkaConfig.toConsumerProperties());
   }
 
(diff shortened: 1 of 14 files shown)


apps/threat-detection/src/main/java/com/akto/threat/detection/tasks/AbstractKafkaConsumerTask.java
```diff
@@ -48,16 +39,14 @@ public void run() {
 
     pollingExecutor.execute(
         () -> {
-          // Poll data from Kafka topic
+          beforePollLoop();
           while (true) {
             ConsumerRecords<String, V> records =
                 kafkaConsumer.poll(
                     Duration.ofMillis(kafkaConfig.getConsumerConfig().getPollDurationMilli()));
-            if (records.isEmpty()) {
```
ayushaga14, line 56: why remove this check?

gauravakto: in case of records empty, an empty list will be sent and every minute we will log 0 records process/min


if remove the check then this log will not print.

TangoBeeAkto (approved):

PR #5004, opened 2026-05-01
feat: add fetchApiSequences endpoint to database-abstractor

gauravakto:
## Summary
- Adds `fetchApiSequences()` action to `DbAction` — fetches all `api_sequences` documents for the account
- Registers `api/fetchApiSequences` GET endpoint in `struts.xml`
- Response field: `apiSequencesList` (matches existing getter/setter on `DbAction`)

This endpoint is consumed by `ClientActor.fetchApiSequences()` in threat-detection's `SequenceCache` for sequence-based anomaly detection.

## Test plan
- [ ] GET `/api/fetchApiSequences` returns 200 with `apiSequencesList` array
- [ ] Returns empty list when no sequences exist (not an error)
- [ ] Returns 422 on DB exception

### DIFF
--- a/apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
+++ b/apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
@@ -774,7 +774,7 @@ public class DbAction extends ActionSupport {
 
     public String fetchApiSequences() {
         try {
-            apiSequencesList = ApiSequencesDao.instance.findAll(new BasicDBObject());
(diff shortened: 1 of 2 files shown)


apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -772,6 +772,16 @@ public String writeApiSequences() {
         return Action.SUCCESS.toUpperCase();
     }
 
+    public String fetchApiSequences() {
```
ayushaga14, line 775: batch?

notshivansh (approved):

PR #5087, opened 2026-05-12
System proxy domains

shubhamakto:


apps/dashboard/src/main/resources/struts.xml
```diff
@@ -678,6 +678,19 @@
...
+            <interceptor-ref name="defaultStack" />
+            <result name="SUCCESS" type="json">
+                <param name="root">deviceDomainConfig</param>
+            </result>
+            <result name="ERROR" type="json">
+                <param name="statusCode">422</param>
+                <param name="ignoreHierarchy">false</param>
+                <param name="includeProperties">^actionErrors.*</param>
+            </result>
+        </action>
```
aanchalakto: ADMIN_ACTIONS should be used if role allowed for this api is ADMIN

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -130,6 +136,24 @@ public String updateCidrList() {
         return Action.SUCCESS.toUpperCase();
     }
 
+    public String updateDeviceDomainListDelta() {
```
aanchalakto: no changes required in this file

apps/database-abstractor/src/main/resources/struts.xml
```diff
@@ -68,6 +68,28 @@
             </result>
         </action>
 
+        <action name="api/updateDeviceDomainListDelta" class="com.akto.action.DbAction" method="updateDeviceDomainListDelta">
```
aanchalakto: no changes required in this file

libs/dao/src/main/java/com/akto/dao/DeviceDomainConfigDao.java
```diff
@@ -0,0 +1,27 @@
...
+    public String getCollName() {
+        return "device_domain_config";
+    }
+
+    @Override
+    public Class<DeviceDomainConfig> getClassT() {
+        return DeviceDomainConfig.class;
+    }
+
+    public void createIndicesIfAbsent() {
```
aanchalakto: Once refer the existing createIndicesIfAbsent of any other dao class. eg - TestCollectionPropertiesDao. The getDBName() passed in createCollectionIfAbsent passes the db name as "common" but i think your collection will be created in all accounts so use String dbName = Context.accountId.get()+"";  common db is for common collections and accountId is for collections specific to all accounts

libs/utils/src/main/java/com/akto/data_actor/DbLayer.java
```diff
@@ -87,6 +89,37 @@ public static void updateApiCollectionName(int vxlanId, String name) {
         );
     }
```
aanchalakto, line 90: no changes required in this class

aanchalakto (approved):

PR #5140, opened 2026-05-19
update issue summary APIs, skip agentic data in API dashboard

notshivansh:


libs/dao/src/main/java/com/akto/dao/testing_run_findings/TestingRunIssuesDao.java
```diff
@@ -326,39 +325,55 @@ public String getFilterKeyString(){
...
+
+    public Bson getSummaryIdsFilterForDashboardContext(CONTEXT_SOURCE contextSource, String filterKey) {
+        if (contextSource == null) {
+            return Filters.empty();
+        }
+        return getSummaryIdsFilterForTestRuns(filterKey, Filters.eq(TestingRun.DASHBOARD_CONTEXT, contextSource));
+    }
+
+    /** Issues linked to test runs with an explicit dashboardContext different from the current dashboard. */
+    private Bson getOtherDashboardContextFilterForIssues(CONTEXT_SOURCE contextSource) {
```
krngrover94, line 368: why this OtherDashboardContextFilter as we are passing only dashboardContext which will be the same i guess

notshivansh: when we get using apicollections, then we want to filter out the ones, not belonging to this context, but also keep the ones, which do not have dashboard context, but have the api collection [ legacy ones, and ones whose test runs have been deleted ].

krngrover94 (approved):

PR #5159, opened 2026-05-21
skip deactovate collection api info

ayushaga14:


libs/utils/src/main/java/com/akto/data_actor/DbLayer.java
```diff
@@ -481,7 +481,11 @@ public static Config.AutomatedAiTestingKeyConfig fetchModelApiKey() {
     }
 
     public static List<ApiInfo> fetchApiInfos() {
-        return ApiInfoDao.instance.findAll(new BasicDBObject(), Projections.exclude(ApiInfo.RATELIMITS));
+        List<ApiInfo> apiInfos = ApiInfoDao.instance.findAll(new BasicDBObject(), Projections.exclude(ApiInfo.RATELIMITS));
+        Set<Integer> deactivated = UsageMetricCalculator.getDeactivated();
```
gauravakto, line 487: where does this fetch from ?

gauravakto: ignore found it. we should once verify the code-path in miniruntime what effect will this have ?

gauravakto (dismissed):

notshivansh (approved):

PR #5179, opened 2026-05-25
add tagsList to ApiInfo and update ApiInfoBulkUpdate for tag handling

harshithb3304:


libs/utils/src/main/java/com/akto/bulk_update_util/ApiInfoBulkUpdate.java
```diff
@@ -52,6 +53,14 @@ public static List<WriteModel<ApiInfo>> getUpdatesForApiInfo(List<ApiInfo> apiIn
...
             subUpdates.add(Updates.set(ApiInfo.LAST_SEEN, apiInfo.getLastSeen()));
 
+            // tagsList — user-agent categories accumulated across traffic
+            List<CollectionTags> tagsList = apiInfo.getTagsList();
+            if (tagsList == null || tagsList.isEmpty()) {
+                subUpdates.add(Updates.setOnInsert(ApiInfo.TAGS_STRING, new ArrayList<>()));
+            } else {
+                subUpdates.add(Updates.addEachToSet(ApiInfo.TAGS_STRING, new ArrayList<>(tagsList)));
+            }
+
```
gauravakto, line 63: test the use case where some tags were existing. what will this api do ? overwrite the old ones or just add the new ones ?? 


Ideal behaviour.
- for existing leave them as is. 
- for exissting with same keyName update the value
- for completely new just add to the list.

gauravakto (approved):

PR #5180, opened 2026-05-25
Feature/store query data cy

Ark2307:


apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -3932,11 +3932,27 @@ public String storeSpans() {
         return Action.SUCCESS.toUpperCase();
     }
 
-    public String storeAgentQueryData() {
+    @Getter @Setter
+    private List<com.akto.utils.elasticsearch.AgentQueryRecord> agentQueryRecords;
+
+    public String storeAgentQueryRecords() {
+        try {
+            com.akto.utils.elasticsearch.ElasticSearchClient.instance().bulkIndexAgentQueryRecords(agentQueryRecords);
```
gauravakto, line 3940: imports to top. 

also why are we using elasticsearch ? For the expected read query patterns do check manually..will Elastic search actually work performant or not.

gauravakto (approved):

PR #5181, opened 2026-05-25
Handling from module data and adding batching

Ark2307:


libs/dao/src/main/java/com/akto/DaoInit.java
```diff
@@ -292,7 +288,7 @@ public static CodecRegistry createCodecRegistry(){
                         nodeClassModel, connectionClassModel, edgeClassModel, replaceDetailClassModel, modifyHostDetailClassModel, fileUploadClassModel
                 ,fileUploadLogClassModel, codeAnalysisCollectionClassModel, codeAnalysisApiLocationClassModel, codeAnalysisApiInfoClassModel, codeAnalysisApiInfoKeyClassModel,
                 riskScoreTestingEndpointsClassModel, OrganizationFlagsClassModel, jobParams, tlsAuthClassModel, digestAuthParamClassModel, collectionTagsModel, mcpReconResultClassModel
-                , mcpReconRequestClassModel, agentConversationResultClassModel, agentTrafficLogClassModel, agentQueryDataClassModel, apiSequencesClassModel).automatic(true).build());
+                , mcpReconRequestClassModel, agentConversationResultClassModel, agentTrafficLogClassModel, apiSequencesClassModel).automatic(true).build());
```
gauravakto, line 291: local testing won't work ? maybe if we left as is on local mongo can be used ?

gauravakto (approved):

gauravakto (approved):

PR #5183, opened 2026-05-26
add API for guardrails service logs

notshivansh:


libs/dao/src/main/java/com/akto/dao/GuardrailsServiceLogsDao.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static final int maxDocuments = 100_000;
+    public static final int sizeInBytes = 100_000_000;
+
+    public static final GuardrailsServiceLogsDao instance = new GuardrailsServiceLogsDao();
+
+    public void createIndicesIfAbsent() {
+        boolean exists = false;
+        String dbName = Context.accountId.get() + "";
+        MongoDatabase db = clients[0].getDatabase(dbName);
```
abhijeet-akto, line 21: Do we need to add null check in case clients don't have any entry?

libs/dao/src/main/java/com/akto/dao/GuardrailsServiceLogsDao.java
```diff
@@ -0,0 +1,49 @@
...
+            }
+        }
+
+        if (!exists) {
+            if (DbMode.allowCappedCollections()) {
+                db.createCollection(getCollName(), new CreateCollectionOptions().capped(true).maxDocuments(maxDocuments).sizeInBytes(sizeInBytes));
+            } else {
+                db.createCollection(getCollName());
+            }
+        }
```
abhijeet-akto: `try {
    if (DbMode.allowCappedCollections()) {
        db.createCollection(
            getCollName(),
            new CreateCollectionOptions()
                .capped(true)
                .maxDocuments(maxDocuments)
                .sizeInBytes(sizeInBytes)
        );
    } else {
        db.createCollection(getCollName());
    }
} catch (MongoCommandException e) {
    // NamespaceExists = collection already exists
    if (e.getErrorCode() != 48) {
        throw e;
    }
}`

We can simplify above code of iteration using this.

abhijeet-akto (dismissed):

abhijeet-akto (approved):

PR #5233, opened 2026-05-29
Feature/agentic category filter

rahul-akto:


notshivansh:
this would decrease visibility as we're silently skipping the tests here.
alternatively, we can skip them, but keep the test run result entries, with a failed message.

notshivansh (approved):

PR #5261, opened 2026-06-02
Fixes/api rate limiting

gauravakto:
## API Hit Count

Tracks how many times each API endpoint is called per minute across all consumers to detect API-level abuse (e.g. a flood of requests to one endpoint across many IPs, each individually under the per-IP rate limit).

### How it works

**Hot path**: Per request, two extra fields (`apiLevelWindow`, `apiLevelThreshold`) are packed into the existing `threat_input_stream` Redis stream message. Config is fetched from `FilterCache.getApiLevelRateLimitRule()` and embedded in the message so the background consumer needs no access to filter state.

**Cold path (Lua)**: The existing `async_threat_process.lua` script handles API hit counting alongside IP distribution:
- `INCR apiCount|{collectionId}|{url}|{method}|{binId}` per request (8h TTL)
- `ZADD apiCountIndex score={binId}` to index active APIs for the relay cron
- Sliding window sum across `apiLevelWindow` bins — if total ≥ `apiLevelThreshold`, returns a breach to Java which pushes a `MaliciousEventKafkaEnvelope` alert

**Relay cron (every 1 min)**: Reads all per-bin counts from Redis via `ZRANGEBYSCORE` + `MGET` and bulk-inserts `ApiHitCountInfo` records to DB for dashboards.

---

### Root cause

The distribution module buckets unique IP+API pairs by request count (b1=1–10, b2=11–50, etc.) using a Count-Min Sketch (CMS). The original logic detected "first time this IP+API is seen in this window" by checking `if count == 1` after querying the CMS.

**CMS is probabilistic and can overestimate.** The error bound is `epsilon × N` where N is total insertions across all keys in the sketch for that minute. With `epsilon=0.01` and just 100 unique IP+API pairs per minute, the overestimate is already ≥ 1 — meaning a new IP+API's very first request returns `count=2` instead of `1`. The `count==1` guard is skipped, `b1` never gets its initial `+1`, but at the b1→b2 boundary (count=11) `b1` gets decremented → **b1 goes negative**.

Confirmed via Redis logs: every first-seen overestimate shows `count=2`, and every negative guard trigger shows `count=11` (the b1→b2 boundary), `val=0` (b1 was never incremented).

### Fix

Replaced the fragile `count==1` / `getBucketLabel(count-1)` inference with a `prevBuckets|{windowSize}|{windowStart}` Redis hash that stores the last known bucket per IP+API pair per window.

```lua
local prevBucket = redis.call('HGET', prevBucketHashKey, ipApiCmsKey)

if prevBucket == false then
    -- first time this IP+API seen in this window
    redis.call('HINCRBY', distKey, newBucket, 1)
elseif prevBucket ~= newBucket then
    redis.call('HINCRBY', distKey, prevBucket, -1)
    redis.call('HINCRBY', distKey, newBucket, 1)
end

redis.call('HSET', prevBucketHashKey, ipApiCmsKey, newBucket)
```

- First-seen detection is now via `HGET` returning `false` (nil) — independent of CMS accuracy
- Transitions always use the actual last stored bucket, not `count-1` arithmetic
- Negative counts are now impossible regardless of CMS overestimation or pod concurrency
- `HGET`+`HSET` within the same atomic Lua script — no race conditions across pods


apps/threat-detection/src/main/java/com/akto/threat/detection/tasks/MaliciousTrafficDetectorTask.java
```diff
@@ -264,7 +258,7 @@ private void processRecord(HttpResponseParam record) throws Exception {
 
     // Use template URL if available, otherwise fall back to static URL
     // This ensures API counts aggregate on template URLs (e.g., /api/users/INTEGER instead of /api/users/123)
-    String urlForAggregation = matchedTemplate != null ? matchedTemplate.getTemplateString() : url;
+    String urlForAggregation = matchedTemplate != null ? matchedTemplate.getTemplateString() : threatDetector.cleanUrl(url);
```
ayushaga14, line 261: can threatDetector.cleanUrl(url) return null?

gauravakto: no

ayushaga14 (approved):

PR #5286, opened 2026-06-03
feat: implement async Kafka producer for CPU-based policy evaluation

krngrover94:
- Added configuration options for async CPU-based LLM evaluation in guardrails-service.
- Introduced a Kafka producer to handle WARN/ALERT policy deferrals.
- Updated settings and provider logic to support Ollama model integration.
- Enhanced request validation to queue async evaluations for specific policies.
- Refactored model configuration overrides for consistent CPU-based processing.

### DIFF
--- a/apps/agent-guard/python-service/container/src/providers.py
+++ b/apps/agent-guard/python-service/container/src/providers.py
@@ -389,7 +389,7 @@ def _build_openai_compatible(model: str, base_url: str) -> Optional[LLMProvider]
     if not api_key and not base_url:
         logger.warning("[Providers] OPENAI_API_KEY not set and no baseUrl; skipping openai")
         return None
-    return OpenAIProvider(api_key, model or DEFAULT_OPENAI_MODEL, base_url=base_url)
+    return OpenAIProvider(api_key, model or settings.OPENAI_MODEL or DEFAULT_OPENAI_MODEL, base_url=base_url)
 
 
 def _build_anthropic(model: str) -> Optional[LLMProvider]:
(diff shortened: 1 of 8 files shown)


apps/agent-guard/python-service/container/src/settings.py
```diff
@@ -49,6 +49,10 @@ def _port_default(cls, v: object) -> object:
     OPENAI_MODEL: str = ""
     OPENAI_COMPATIBLE_BASE_URL: str = ""
 
+    # ── Ollama (CPU-based async evaluation) ──────────────────────────────
+    OLLAMA_BASE_URL: str = ""
+    OLLAMA_MODEL: str = "gemma4:e4b"
```
abhijeet-akto: use instruct model from gemma family as that is more fined tuned for agent conversations. The default model behaves as autocomplete and that can cause issue.

apps/agent-guard/python-service/container/src/providers.py
```diff
@@ -503,7 +503,8 @@ def build_provider_from_config(entry: Dict[str, Any]) -> Optional[LLMProvider]:
     name = (entry.get("provider") or "").strip().lower()
     model = (entry.get("model") or "").strip()
     if name in ("openai", "ollama", "openai_compatible"):
-        # Entry-level baseUrl wins, then OPENAI_COMPATIBLE_BASE_URL, then default.
-        base_url = (entry.get("baseUrl") or "").strip() or settings.OPENAI_COMPATIBLE_BASE_URL
+        # Entry-level baseUrl wins, then OLLAMA_BASE_URL, then OPENAI_COMPATIBLE_BASE_URL.
+        base_url = (entry.get("baseUrl") or "").strip() or settings.OLLAMA_BASE_URL or settings.OPENAI_COMPATIBLE_BASE_URL
+        model = model or settings.OLLAMA_MODEL
```
abhijeet-akto: This code is not getting used, the code that is getting used is in worker-py

apps/agent-guard/python-service/container/src/settings.py
abhijeet-akto: This code is not getting used, the code that is getting used is in worker-py

PR #5372, opened 2026-06-15
Fix/config fixes

harshithb3304:


apps/threat-detection-backend/src/main/java/com/akto/threat/backend/cron/ConfigRiskSyncCron.java
```diff
@@ -0,0 +1,178 @@
...
+    private static final LoggerMaker loggerMaker = new LoggerMaker(ConfigRiskSyncCron.class, LogDb.THREAT_DETECTION);
+
+    public void setUp() {
+        scheduler.scheduleAtFixedRate(new Runnable() {
+            public void run() {
+                AccountTask.instance.executeTask(new Consumer<Account>() {
+                    @Override
+                    public void accept(Account t) {
+                        try {
+                            syncConfigRiskForAccount(t.getId());
```
notshivansh, line 53: behind stigg, else will execute for all accounts.

notshivansh (approved):

PR #5377, opened 2026-06-15
Import API endpoints from Wiz

orenakto:


apps/api-runtime/src/main/java/com/akto/parsers/HttpCallParser.java
```diff
@@ -162,7 +163,7 @@ public int createCollectionSimple(int vxlanId, boolean isMcpRequest) {
     }
 
 
-    public int createCollectionBasedOnHostName(int id, String host, boolean isMcpRequest)  throws Exception {
+    public int createCollectionBasedOnHostName(int id, String host, boolean isMcpRequest, boolean isWizRequest)  throws Exception {
         FindOneAndUpdateOptions updateOptions = new FindOneAndUpdateOptions();
```
ayushaga14, line 167: why are we adding such specific params? will we keep adding isXyzRequest etc later for other flows too?

apps/api-runtime/src/main/java/com/akto/parsers/HttpCallParser.java
```diff
@@ -511,11 +515,12 @@ public int createApiCollectionId(HttpResponseParams httpResponseParam){
             if (hostNameToIdMap.containsKey(key)) {
                 apiCollectionId = hostNameToIdMap.get(key);
                 updateMcpServerTag(apiCollectionId, isMcpRequest);
+                updateCollectionTag(apiCollectionId, isWizRequest, getWizCollectionTag());
             } else {
                 int id = hostName.hashCode();
                 try {
 
-                    apiCollectionId = createCollectionBasedOnHostName(id, hostName, isMcpRequest);
+                    apiCollectionId = createCollectionBasedOnHostName(id, hostName, isMcpRequest, isWizRequest);
```
ayushaga14: this looks ugly, passing multiple vars like this

apps/api-runtime/src/main/java/com/akto/parsers/HttpCallParser.java
```diff
@@ -568,6 +573,33 @@ private void updateMcpServerTag(int apiCollectionId, boolean isMcpRequest) {
         );
     }
 
+    private void updateCollectionTag(int apiCollectionId, boolean shouldApply, CollectionTags tag) {
```
ayushaga14: why did you write a new function? was there no existing update func, there should be 1

ayushaga14 (changes requested):

ayushaga14 (approved):

PR #5392, opened 2026-06-16
feat: add NHI governance — identities, policies, violations dashboard

TangoBeeAkto:


apps/dashboard/src/main/java/com/akto/utils/crons/Crons.java
```diff
@@ -207,5 +211,48 @@ private static long calculateInitialDelay() {
...
+
+    public void evaluateNhiPoliciesScheduler() {
+        scheduler.scheduleAtFixedRate(new Runnable() {
+            public void run() {
+                AccountTask.instance.executeTask(new Consumer<Account>() {
+                    @Override
+                    public void accept(Account t) {
+                        try {
+                            List<NhiIdentity> identities = NhiIdentityDao.instance.findAll(
+                                com.mongodb.client.model.Filters.eq(NhiIdentity.STATUS, "ACTIVE")
```
notshivansh: no inline imports

apps/database-abstractor/src/main/java/com/akto/action/NhiUpsertAction.java
```diff
@@ -0,0 +1,67 @@
...
+    private List<NhiIdentity> nhiIdentities;
+
+    @Getter
+    private int upsertedCount;
+
+    @Getter
+    private int skippedCount;
+
+    @Getter
+    private boolean success = false;
```
notshivansh: why do we need this ?

libs/dao/src/main/java/com/akto/dto/nhi_governance/NhiPolicy.java
```diff
@@ -32,22 +32,21 @@ public class NhiPolicy {
     private String policyName;
     private String description;
     private String status;
-    private String contextSource;
```
notshivansh, line 35: we're deleting an attribute.
Check if it will affect the existing data, which has this attribute.

libs/utils/src/main/java/com/akto/utils/nhi_governance/NhiIdentityUpsertService.java
```diff
@@ -0,0 +1,171 @@
...
+import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.Updates;
+import org.apache.commons.lang3.StringUtils;
+import org.bson.conversions.Bson;
+import org.bson.types.ObjectId;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class NhiIdentityUpsertService {
```
notshivansh: Use standard mongo commands, ref. MCollection , in the dao file. Any specific use case for the service ?

notshivansh (dismissed):

notshivansh (approved):

PR #5398, opened 2026-06-16
increase timeout

notshivansh:


Ark2307 (approved):

notshivansh:
Contains 2 fixes

1. multiple timed out results created when kafka consumer is timed out.
2. parallel testing for agentic tests. we need to shallow copy the test configurations, so that each test can have their own conversations array.

PR #5434, opened 2026-06-18
cron + backend code for guardrails-compliance

rohan-akto:


apps/dashboard/src/main/java/com/akto/listener/InitializerListener.java
```diff
@@ -3773,11 +3775,17 @@ public void run() {
...
                     processRemedationFilesZip(testingTemplates);
                     processComplianceInfosFromZip(testingTemplates);
-                    
+
                 } catch (Exception e) {
                     e.printStackTrace();
                     logger.errorAndAddToDb("Unable to import remediations", LogDb.DASHBOARD);
                 }
+                try {
+                    processGuardrailComplianceInfosFromZip(testingTemplates);
```
notshivansh, line 3784: make sure this function is called only once, not per account, since we're adding in common collection

PR #5498, opened 2026-06-23
Add accurate, verified vulnerability evidence for test results

shivam-rawat-akto:
Replace LLM-only evidence with a layered pipeline so the Evidence panel and Monaco highlights always agree and never cite hallucinated or noisy values.

Frontend:
- Extract shared locator + credential denylist into vulnerabilityEvidence.js
- Add buildRenderedText() so verification uses the same text as the editor
- Validate segments in SampleDataList (locatable + denylisted) before render
- Fix computeRequestDiff: require a real baseline and skip transport/proxy/CDN headers so broken-auth tests are not flooded with fake "injected" evidence
- Gate the Evidence banner to @akto.io users via func.isAktoUser()

Backend:
- Add EvidenceRouter (category -> detector + prompt hints)
- Add VulnerabilityEvidenceDetector (EDE data-type scan, request diff, CORS headers, broken-auth response status)
- Merge deterministic + LLM segments in AnalyzeVulnerability (deterministic wins); skip LLM when strong deterministic evidence exists
- Harden LLM prompt: verbatim phrases only, no JWT/base64 decoding, no caller credential citations
- Bump vulnerability analysis cache key to v5

### DIFF
--- a/apps/dashboard/src/main/java/com/akto/action/gpt/handlers/EvidenceRouter.java
+++ b/apps/dashboard/src/main/java/com/akto/action/gpt/handlers/EvidenceRouter.java
@@ -12,12 +12,10 @@ import java.util.List;
 public class EvidenceRouter {
 
     public enum Detector {
-        DATATYPE_RESPONSE_SCAN, // sensitive-data exposure: scan response for the data type
-        REQUEST_DIFF,           // injection/fuzzing/authz: the values the test changed
-        RESPONSE_HEADERS,       // CORS / security misconfig: a response header
+        RESPONSE_HEADERS,       // CORS / security misconfig: a precise response header
         RESPONSE_STATUS,        // broken auth / no-auth: 2xx returned after token removed
         CROSS_ATTEMPT,          // rate-limit / multi-account: proof spans attempts
-        LLM_SEMANTIC            // error/stack-trace disclosure, business logic
+        LLM_SEMANTIC            // everything else: LLM selects evidence, FE verifies it
     }
 
     public static class Route {
@@ -42,13 +40,22 @@ public class EvidenceRouter {
     public static Route route(String category) {
         String c = category == null ? "" : category.toUpperCase();
 
+        // Sensitive-data exposure: the data type is already known from the test
+        // sub-type (see edeDataType) and may surface in EITHER the response or the
+        // request (the test's own conditions decide where). We do not re-scan/re-
+        // classify here - the LLM is told the exact data type and asked to copy the
+        // value verbatim; the frontend then verifies it is actually highlightable.
         if (c.startsWith(EDE_PREFIX) || c.contains("EXCESSIVE_DATA") || c.equals("EDE")) {
             return new Route(
-                list("RESPONSE"),
-                list(Detector.DATATYPE_RESPONSE_SCAN),
-                "Evidence is the exposed value in the RESPONSE body or headers.");
+                list("RESPONSE", "REQUEST"),
+                list(Detector.LLM_SEMANTIC),
+                "Evidence is the exposed value of the data type named by this test, shown verbatim in the RESPONSE (or the REQUEST if that is where the sensitive value appears).");
         }
-        if (c.contains("CORS") || c.contains("SECURITY_HEADER") || c.contains("MISCONFIG") || c.contains("SSRF")) {
+        // Header-based misconfig ONLY (CORS / security headers). Keep this first
+        // check precise: SSRF is NOT a response-header finding (its proof is the
+        // injected URL in the request + the fetched content in the response), so it
+        // must not be routed here - it falls through to the request/response path.
+        if (c.contains("CORS") || c.contains("SECURITY_HEADER")) {
             return new Route(
                 list("RESPONSE"),
                 list(Detector.RESPONSE_HEADERS, Detector.LLM_SEMANTIC),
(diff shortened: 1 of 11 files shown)


apps/dashboard/src/main/java/com/akto/action/gpt/handlers/VulnerabilityEvidenceDetector.java
```diff
@@ -0,0 +1,256 @@
...
+    private static final int MAX_PHRASE_CHARS = 80;
+
+    public static BasicDBList detect(BasicDBObject testData) {
+        BasicDBList segments = new BasicDBList();
+        if (testData == null) {
+            return segments;
+        }
+        try {
+            BasicDBObject testContext = (BasicDBObject) testData.get("testContext");
+            BasicDBObject response = (BasicDBObject) testData.get("response");
```
Ark2307: already did this parsing in analyzeVulnerability.java, directly pass this in arguments

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/EvidenceRouter.java
```diff
@@ -0,0 +1,95 @@
...
+            return new Route(
+                list("REQUEST", "RESPONSE"),
+                list(Detector.CROSS_ATTEMPT, Detector.LLM_SEMANTIC),
+                "Evidence spans attempts; cite the RESPONSE status that shows the limit was not enforced.");
+        }
+        // Broken authentication / no-auth: the test removes or breaks the auth token
+        // and the server still returns 2xx. The evidence is response-side (the status
+        // + any data returned), NOT the request headers, so do not run request-diff.
+        if (c.contains("NO_AUTH") || c.contains("ANY_USER") || c.contains("REMOVE_TOKEN")
+                || c.contains("REPLACE_AUTH") || c.contains("ADD_USER_TOKEN") || c.contains("AUTH_BYPASS")
```
Ark2307, line 89: the category NO_AUTH, contains sqli and command injection attacks also, not completely related to the auth token was removed/invalidated, verify this

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/EvidenceRouter.java
```diff
@@ -0,0 +1,95 @@
...
+    public static Route route(String category) {
+        String c = category == null ? "" : category.toUpperCase();
+
+        if (c.startsWith(EDE_PREFIX) || c.contains("EXCESSIVE_DATA") || c.equals("EDE")) {
+            return new Route(
+                list("RESPONSE"),
+                list(Detector.DATATYPE_RESPONSE_SCAN),
+                "Evidence is the exposed value in the RESPONSE body or headers.");
+        }
+        if (c.contains("CORS") || c.contains("SECURITY_HEADER") || c.contains("MISCONFIG") || c.contains("SSRF")) {
```
Ark2307: not true for ssrf, since this is the first check for determination, it should be highly precise

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/EvidenceRouter.java
```diff
@@ -0,0 +1,95 @@
...
+                list("RESPONSE"),
+                list(Detector.RESPONSE_STATUS, Detector.LLM_SEMANTIC),
+                "The auth token was removed/invalidated; evidence is the 2xx RESPONSE status and any data the response returned. Do NOT cite the request's own auth token or headers.");
+        }
+        if (c.contains("BOLA") || c.contains("BFLA") || c.contains("IDOR") || c.contains("AUTH")
+                || c.contains("ACCESS") || c.contains("PRIVILEGE") || c.contains("ACCOUNT")) {
+            return new Route(
+                list("REQUEST", "RESPONSE"),
+                list(Detector.REQUEST_DIFF, Detector.LLM_SEMANTIC),
+                "Evidence is the changed request object id/identity AND the RESPONSE data wrongly returned. Do NOT cite the caller's auth token.");
```
Ark2307: the test validation relies on percentage match schema for vulnerability, so the response data is returned right but it shouldn't have in ideal case, wrong semantic language

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/VulnerabilityEvidenceDetector.java
```diff
@@ -0,0 +1,256 @@
...
+            String category = testContext != null ? testContext.getString("category") : null;
+            String cUpper = category == null ? "" : category.toUpperCase();
+
+            EvidenceRouter.Route route = EvidenceRouter.route(category);
+
+            if (route.detectors.contains(EvidenceRouter.Detector.DATATYPE_RESPONSE_SCAN)
+                    && response != null && cUpper.startsWith(EvidenceRouter.EDE_PREFIX)) {
```
Ark2307: response != null && cUpper.startsWith(EvidenceRouter.EDE_PREFIX, redundant checks, already checked in EvidenceRouter.route(category);

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/EvidenceRouter.java
```diff
@@ -0,0 +1,95 @@
...
+    private static <T> List<T> list(T... items) {
+        return Arrays.asList(items);
+    }
+
+    public static Route route(String category) {
+        String c = category == null ? "" : category.toUpperCase();
+
+        if (c.startsWith(EDE_PREFIX) || c.contains("EXCESSIVE_DATA") || c.equals("EDE")) {
+            return new Route(
+                list("RESPONSE"),
```
Ark2307: the EDE tests depends on the conditions to detect the data type, not necessarily in response, what will happen in that case?

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/VulnerabilityEvidenceDetector.java
```diff
@@ -0,0 +1,256 @@
...
+                String typeName = cUpper.substring(EvidenceRouter.EDE_PREFIX.length());
+                addDataTypeResponseSegments(segments, response, typeName);
```
Ark2307: why do you need to scan it, it is already vulnerable we know that and we have already parsed the sample for the datatype, why do we need this, construct a definitive system prompt for ai and get datatypes from db if need be?

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/VulnerabilityEvidenceDetector.java
```diff
@@ -0,0 +1,256 @@
...
+            }
+
+            if (route.detectors.contains(EvidenceRouter.Detector.REQUEST_DIFF)) {
+                addRequestDiffSegments(segments, testData.get("modifiedRequestValues"));
+            }
+
+            if (route.detectors.contains(EvidenceRouter.Detector.RESPONSE_STATUS) && response != null) {
+                addResponseStatusSegment(segments, response);
```
Ark2307, line 35: here the bua category will come where command injection tests would be there too, handle those

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/VulnerabilityEvidenceDetector.java
```diff
@@ -0,0 +1,256 @@
...
+            String phrase = modified.toString();
+            if (phrase.trim().isEmpty()) {
+                continue;
+            }
+            String reason = (original != null)
+                    ? "The request changed \"" + field + "\" to \"" + truncate(phrase) + "\" and the endpoint accepted it."
+                    : "The request injected \"" + truncate(phrase) + "\" into \"" + field + "\" and the endpoint accepted it.";
+            segments.add(makeSegment("REQUEST", field, phrase, reason));
+            if (++count >= 15) {
+                break;
```
Ark2307: this check is vague, we modify request headers a lot, and what of advanceconfigurations where user deliberately change requests as precompute to run tests like deleting cf-headers

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/EvidenceRouter.java
shivam-rawat-akto: Handled this, 

Now sqli and command ijection attacks are checked before, and if none of them satisfy, we check NO_AUTH

apps/dashboard/src/main/java/com/akto/action/gpt/handlers/VulnerabilityEvidenceDetector.java
shivam-rawat-akto: it has the same root cause, 
now command injection tests are routed to LLM_SYMANTIC, 
and here it comes only in the case of bua.

Ark2307 (approved):

PR #5499, opened 2026-06-23
parse cookies before subtype finding

gauravakto:
## Summary

- Previously, cookie redaction in the non-`redactAll` path was calling `redactCookie()` which redacted **all** cookie values unconditionally, regardless of their subtype
- Added `handleCookies()` which parses the cookie header, checks each cookie's subtype individually, and only redacts cookies whose subtype is flagged as sensitive
- Replaced the old cookie-handling logic in `handleHeaders()` with a call to `handleCookies()` for selective, per-cookie redaction

## Test plan

- [x] Verify that cookies with sensitive subtypes (e.g. auth tokens) are redacted
- [x] Verify that non-sensitive cookies are preserved as-is
- [x] Verify that the `redactAll=true` path still redacts all cookies via `redactCookie()`

### DIFF
--- a/apps/mini-runtime/src/main/java/com/akto/utils/RedactSampleData.java
+++ b/apps/mini-runtime/src/main/java/com/akto/utils/RedactSampleData.java
@@ -94,7 +94,7 @@ public class RedactSampleData {
         return finalUrl;
     }
 
-    private static void handleCookies(List<String> cookieList, Map<String, List<String>> responseHeaders) {
+   private static void handleCookies(List<String> cookieList, Map<String, List<String>> responseHeaders) {
         Map<String, String> cookieMap = AuthPolicy.parseCookie(cookieList);
         StringBuilder newCookie = new StringBuilder();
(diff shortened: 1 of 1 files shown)


apps/mini-runtime/src/main/java/com/akto/utils/RedactSampleData.java
```diff
@@ -94,6 +94,24 @@ private static String handleQueryParams(String url, boolean redactAll) {
...
+            String cookieName = entry.getKey();
+            String cookieValue = entry.getValue();
+            SingleTypeInfo.SubType subType = KeyTypes.findSubType(cookieValue, cookieName, null);
+            newCookie.append(cookieName).append("=");
+            if (SingleTypeInfo.isRedacted(subType.getName())) {
+                newCookie.append(redact(cookieValue));
+            } else {
+                newCookie.append(cookieValue);
+            }
+            newCookie.append(";");
```
ayushaga14: this would leave ; at the end of every cookie val

notshivansh (approved):

PR #5584, opened 2026-06-29
AWS service graph support

aanchalakto:


libs/utils/src/main/java/com/akto/tracing/bedrock/BedrockAgentTraceParser.java
```diff
@@ -0,0 +1,321 @@
...
+            JsonNode root = OBJECT_MAPPER.readTree(jsonStr);
+
+            // Check for awsMetadata field
+            JsonNode awsMetadata = root.path("awsMetadata");
+            if (awsMetadata.isMissingNode()) {
+                return false;
+            }
+
+            // Check for either harness or bedrock execution role
+            boolean hasHarnessRole = awsMetadata.has("harness-execution-role");
```
krngrover94, line 39: why we need to check roles?

krngrover94 (approved):

PR #5595, opened 2026-06-29
Support testRunResultSummaryId on testing log inserts.

shivam-rawat-akto:
Extend the Log model and database-abstractor insertTestingLog path so hybrid testing modules can persist logs scoped to a specific test run result summary.

### DIFF
--- a/libs/utils/src/main/java/com/akto/log/LoggerMaker.java
+++ b/libs/utils/src/main/java/com/akto/log/LoggerMaker.java
@@ -296,11 +296,23 @@ public class LoggerMaker  {
         return true;
     }
     
+    /**
+     * Stamps the current thread's activity context onto the log only when both the type and a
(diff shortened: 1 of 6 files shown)


libs/utils/src/main/java/com/akto/log/LoggerMaker.java
```diff
@@ -299,7 +299,9 @@ private Boolean checkUpdate(){
     private void insert(String info, String key, LogDb db) {
         String text = aClass + " : " + info;
         Log log = new Log(text, key, Context.now());
-        
+        log.setActivityId(Context.activityId.get());
+        log.setActivityType(Context.activityType.get());
```
notshivansh: null checks

notshivansh (dismissed):

notshivansh (approved):

PR #5596, opened 2026-06-29
Stamp testing logs with testRunResultSummaryId during test execution.

shivam-rawat-akto:
Use a ThreadLocal context to tag LoggerMaker output per run, persist transient testLogs to the logs collection, and carry summary id over the hybrid insertTestingLog API.

### DIFF
--- a/apps/mini-testing/src/main/java/com/akto/testing/TestExecutor.java
+++ b/apps/mini-testing/src/main/java/com/akto/testing/TestExecutor.java
@@ -899,6 +899,25 @@ public class TestExecutor {
         return respMap;
     }
 
+    /**
+     * Establishes a clean activity context for a test run. Tests start from several flows that run on
+     * pooled/reused threads, so we always set explicitly when a summaryId is present and otherwise clear
+     * any stale value left behind by a previous task on the same thread.
+     */
+    public static void setTestRunActivityContext(ObjectId summaryId) {
+        if (summaryId != null) {
+            Context.activityId.set(summaryId.toHexString());
+            Context.activityType.set(Log.ActivityType.TESTING_RUN_RESULT_SUMMARY_ACTIVITY);
+        } else {
+            clearActivityContext();
+        }
+    }
+
+    public static void clearActivityContext() {
+        Context.activityId.remove();
+        Context.activityType.remove();
+    }
+
     public Void startWithLatch(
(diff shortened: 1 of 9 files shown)


apps/mini-testing/src/main/java/com/akto/testing/kafka_utils/ConsumerUtil.java
```diff
@@ -70,25 +71,36 @@ public static SingleTestPayload parseTestMessage(String message) {
...
+                executor.insertResultsAndMakeIssues(Collections.singletonList(runResult), singleTestPayload.getTestingRunResultSummaryId());
+
+                testedApisMap.put(apiInfoKey, Context.now());
+
+                loggerMaker.insertImportantTestingLog("Test completed for: " + apiInfoKey + " with subcategory: " + subCategory + " in " + (Context.now() - timeNow) + " seconds");
+            }
+        } finally {
+            Context.activityId.remove();
+            Context.activityType.remove();
         }
```
notshivansh, line 99: as a safety check, clear these values, before any test starts

apps/mini-testing/src/main/java/com/akto/testing/TestExecutor.java
```diff
@@ -920,6 +927,9 @@ public Void startWithLatch(
             }
         } catch (Exception e) {
             loggerMaker.errorAndAddToDb(e, "error while running tests: " + e);
+        } finally {
+            Context.activityId.set(previousActivityId);
+            Context.activityType.set(previousActivityType);
```
notshivansh: how is the previous activity id used.

Testing is single process, i.e. at any point of time, only one test can execute on it.

libs/utils/src/main/java/com/akto/log/LoggerMaker.java
```diff
@@ -300,7 +314,9 @@ private void insert(String info, String key, LogDb db) {
         }
 
         Log log = new Log(text, key, Context.now());
-        
+        log.setActivityId(Context.activityId.get());
+        log.setActivityType(Context.activityType.get());
```
notshivansh: null checks and only set for specific activity types.

Also, check if these extra fields are handled in the logging API, which sends data to cyborg.

notshivansh (approved):

PR #5597, opened 2026-06-29
Add per-run testing logs on the test result page.

shivam-rawat-akto:
Stamp logs with testRunResultSummaryId, expose a summary-scoped fetch API, and show a Logs section on SingleTestRunPage so each run's service logs are viewable in the dashboard.

### DIFF
--- a/libs/dao/src/main/java/com/akto/DaoInit.java
+++ b/libs/dao/src/main/java/com/akto/DaoInit.java
@@ -467,7 +467,8 @@ public class DaoInit {
                 new EnumCodec<>(GlobalEnums.DashboardCategory.class),
                 new EnumCodec<>(McpRegistryConfig.RegistryType.class),
                 new EnumCodec<>(McpAllowlist.Source.class),
-                new EnumCodec<>(GuardrailPolicies.ModelRole.class)
+                new EnumCodec<>(GuardrailPolicies.ModelRole.class),
+                new EnumCodec<>(Log.ActivityType.class)
         );
 
         return fromRegistries(MongoClientSettings.getDefaultCodecRegistry(), pojoCodecRegistry,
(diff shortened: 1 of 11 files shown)


libs/dao/src/main/java/com/akto/dao/LogsDao.java
```diff
@@ -34,6 +34,9 @@ public void createIndicesIfAbsent() {
 
         String[] fieldNames = {Log.TIMESTAMP};
         MCollection.createIndexIfAbsent(getDBName(), getCollName(), fieldNames,false);
+
+        String[] activityFields = {Log.ACTIVITY_TYPE, Log.ACTIVITY_ID};
+        MCollection.createIndexIfAbsent(getDBName(), getCollName(), activityFields,false);
     }
```
notshivansh: add timestamp as well.

notshivansh (dismissed):

libs/utils/src/main/java/com/akto/log/LoggerMaker.java
```diff
@@ -331,6 +340,8 @@ private void insert(String info, String key, LogDb db) {
 
         String text = aClass + " : " + info;
         Log log = new Log(text, key, Context.now());
+        log.setActivityId(Context.activityId.get());
+        log.setActivityType(Context.activityType.get());
```
notshivansh: add null checks, and only set for the acitivity types, we know.

shivam-rawat-akto: At the moment there's just 1 activity type, 
will add null check

libs/dao/src/main/java/com/akto/DaoInit.java
```diff
@@ -467,7 +467,8 @@ public static CodecRegistry createCodecRegistry(){
                 new EnumCodec<>(GlobalEnums.DashboardCategory.class),
                 new EnumCodec<>(McpRegistryConfig.RegistryType.class),
                 new EnumCodec<>(McpAllowlist.Source.class),
-                new EnumCodec<>(GuardrailPolicies.ModelRole.class)
+                new EnumCodec<>(GuardrailPolicies.ModelRole.class),
+                new EnumCodec<>(Log.ActivityType.class)
```
notshivansh, line 471: we store log activity as string.
we shouldn't need to register the enum in dao, in that case

notshivansh (approved):

PR #5686, opened 2026-07-07
otel ingestion service

notshivansh:


notshivansh:
Gaps:


- cross batch merging of events
- matching only on prompt id. which is a UUID, but could be the same for two users across different sessions

abhijeet-akto (approved):

PR #5711, opened 2026-07-09
add support for detecting statistical anomaly for tool calls and errors

ayushaga14:


apps/guardrails-service/container/src/pkg/session/anomaly.go
```diff
@@ -0,0 +1,158 @@
+package session
+
+import (
+	"context"
+	"fmt"
```
abhijeet-akto, line 5: Move this logic to akto-gateway so that we don't have to re-create redis-cli

apps/guardrails-service/container/src/pkg/validator/service.go
```diff
@@ -726,8 +761,115 @@ func extractHostHeader(headers map[string]string) string {
...
+			}
+			if cfg.Behaviour == "block" {
+				return &mcp.ValidationResult{
+					Allowed:   false,
+					Reason:    details,
+					Behaviour: "block",
+					Metadata: types.ThreatMetadata{
+						PolicyName:   cfg.PolicyName,
+						RuleViolated: session.AnomalyTypeToolCallRate,
+						Severity:     "MEDIUM",
```
abhijeet-akto: severity should be from policy instead of hard-coding it to MEDIUM.

abhijeet-akto (dismissed):

abhijeet-akto (approved):

PR #5717, opened 2026-07-09
Implement pagination for guardrail policies API and UI

nayanakto:


apps/dashboard/src/main/java/com/akto/action/GuardrailPoliciesAction.java
```diff
@@ -66,6 +66,11 @@ public class GuardrailPoliciesAction extends UserAction {
     @Setter
     private List<String> policyIds;
 
+    @Setter
+    private Integer skip;
+    @Setter
```
Ark2307, line 71: use primitive int

Ark2307 (approved):

PR #5739, opened 2026-07-10
Set module type for DataActor in Main and ClientActor

TangoBeeAkto:


Ark2307 (dismissed):

libs/utils/src/main/java/com/akto/data_actor/ClientActor.java
```diff
@@ -3671,10 +3674,112 @@ public static Map<String, List<String>> buildHeaders() {
...
+            connection.setDoOutput(true);
+            try (java.io.OutputStream os = connection.getOutputStream()) {
+                os.write(payload);
+            }
+
+            int status = connection.getResponseCode();
+            if (status != 200) {
+                loggerMaker.warn("Token exchange for moduleType " + moduleType + " not applied (status " + status + "); continuing with raw token");
+                return;
+            }
```
notshivansh: not the standard way to make HTTP requests.
check ApiExecutor.

libs/utils/src/main/java/com/akto/data_actor/ClientActor.java
```diff
@@ -3671,10 +3674,112 @@ public static Map<String, List<String>> buildHeaders() {
...
+
     public static String getAuthToken() {
+        ensureModuleTypeExchanged();
+        if (exchangedAuthToken != null) {
+            return exchangedAuthToken;
+        }
         return System.getenv("DATABASE_ABSTRACTOR_SERVICE_TOKEN");
     }
 
+    private static void ensureModuleTypeExchanged() {
```
notshivansh, line 3696: add retries for exchanges.
we use retries for other clientActor calls as well.
search for call with backoff

notshivansh (approved):

PR #5741, opened 2026-07-10
Implement module type handling and API enhancements for guardrail das…

TangoBeeAkto:
…hboard

- Added support for module types in QuickStartAction and related API endpoints.
- Updated struts.xml to include new action for fetching module types.
- Enhanced API calls in frontend components to handle scope selection.
- Integrated module type initialization in main application classes.
- Improved token exchange logic in ClientActor and related classes.

### DIFF
--- a/apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/quick_start/components/HybridSaasSource.jsx
+++ b/apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/quick_start/components/HybridSaasSource.jsx
@@ -14,7 +14,9 @@ function HybridSaasSource() {
     const copyCommandUtil = (data)=>{func.copyToClipboard(data, ref, null)}
 
     const [selectedExpiryDuration, setSelectedExpiryDuration] = useState(6);
-    const [selectedScope, setSelectedScope] = useState([]);
+    // This page is specifically for the hybrid SaaS runtime helm chart, so default the
+    // generated token's scope to the runtime module rather than leaving it unscoped.
+    const [selectedScope, setSelectedScope] = useState(['MINI_RUNTIME']);
     const [scopeOptions, setScopeOptions] = useState([]);
 
     const fetchRuntimeHelmCommand = async(selectedExpiryDuration, selectedScope) => {
--- a/apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/quick_start/components/HybridSaasSource.jsx
+++ b/apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/quick_start/components/HybridSaasSource.jsx
@@ -8,7 +8,7 @@ import InformationBannerComponent from './shared/InformationBannerComponent'
 
 // This page is specifically for the hybrid SaaS runtime helm chart, so the token it
 // generates is always scoped to the runtime module. Token generation for other scopes
-// (guardrails, threat-detection, etc.) has its own separate card - see databaseAbstractorTokenObj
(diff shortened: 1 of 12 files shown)


apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/quick_start/components/HybridSaasSource.jsx
```diff
@@ -53,11 +65,23 @@ function HybridSaasSource() {
                 menuItems={expiryDurationOptions}
                 value={getLabelFromValue(selectedExpiryDuration)}
                 initial={selectedExpiryDuration}
-                selected={(type) => {setSelectedExpiryDuration(type); fetchRuntimeHelmCommand(type)}}
+                selected={(type) => {setSelectedExpiryDuration(type); fetchRuntimeHelmCommand(type, selectedScope)}}
```
notshivansh: this page is for the helm chart. 

since we're using this token for a lot of other use cases, apart from the helm chart, it should deserve its own screen or page, rather than being clubbed here, maybe in settings

notshivansh: And the token we generate here, should be default attached to runtime, since that's what the hybrid saas helm chart is for.

libs/dao/src/main/java/com/akto/dto/monitoring/ModuleInfo.java
```diff
@@ -111,7 +111,8 @@ public enum ModuleType {
         TRAFFIC_COLLECTOR,
         AKTO_AGENT_GATEWAY,
         DAST,
-        AGENTIC_TESTING
+        AGENTIC_TESTING,
+        GUARDRAIL
```
notshivansh: we have couple of modules with guardrails...

leave a comment here that this is `guardrails-service`

notshivansh (dismissed):

apps/guardrails-service/container/src/pkg/auth/jwt.go
```diff
@@ -1,20 +1,103 @@
...
+func InitModuleType(logger *zap.Logger, moduleType string) {
+	rawToken := os.Getenv("DATABASE_ABSTRACTOR_SERVICE_TOKEN")
+	if rawToken == "" {
+		return
+	}
+
+	baseURL := os.Getenv("DATABASE_ABSTRACTOR_SERVICE_URL")
+	if baseURL == "" {
+		baseURL = "https://ultron.akto.io"
+	}
```
notshivansh: I think these env variables are already referenced, in guardrails service.
use the same ones, or let them come from a function.

libs/utils/src/main/java/com/akto/data_actor/ClientActor.java
```diff
@@ -1357,16 +1357,125 @@ public List<CodeAnalysisRepo> findReposToRun()  {
...
+            String rawToken = getRawAuthToken();
+            if (rawToken == null) {
+                return;
+            }
+
+            BasicDBObject obj = new BasicDBObject();
+            obj.put("moduleType", moduleType);
+            byte[] payload = obj.toString().getBytes(java.nio.charset.StandardCharsets.UTF_8);
+
+            connection = (java.net.HttpURLConnection) new java.net.URL(url + "/exchangeToken").openConnection();
```
notshivansh: not needed changes here.
or do the same fixes here as well.

notshivansh (approved):

PR #5798, opened 2026-07-15
fix: show Malicious tag for skills with malicious_skill_detected events

shivam-rawat-akto:
Derive isMalicious from live violation events on Agentic Assets list load, so skills with malicious detections in details get the list marker without relying only on the cron-written ApiInfo tag. Keep tag enrichment as a case-insensitive backup.


apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/observe/agentic/agenticObserveApi.js
```diff
@@ -113,6 +113,26 @@ export function aggregateViolationsByCollectionId(violationRows = [], collection
     return byCollection;
 }
 
+// Skill identity lives on the event URL as /skills/<skillName> (not on host).
+export function skillNameFromUrl(url) {
+    if (!url) return null;
+    const idx = url.indexOf("skills/");
```
Ark2307, line 119: this code is already written to find asset, check in violations page

PR #5845, opened 2026-07-21
Fix/db abs img vulns

ayushaga14:


pom.xml
```diff
@@ -16,7 +16,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <sonar.organization>akto-api-security</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
```
notshivansh: not needed. we can safely remove sonar

pom.xml
```diff
@@ -25,6 +25,56 @@
...
+            <dependency>
+                <groupId>io.netty</groupId>
+                <artifactId>netty-bom</artifactId>
+                <version>4.1.135.Final</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <!-- Pin transitive libs to fixed versions to clear Grype Critical/High
+                 CVEs across all modules/images (leaked in via shared libs & SDKs). -->
+            <dependency>
```
notshivansh, line 39: these versions are pinned here.
remove all direct and transitive occurences for these deps

pom.xml
```diff
@@ -98,6 +148,13 @@
                     <append>true</append>
 				</configuration>
             </plugin>                 
+            <plugin>
+                <!-- Modern surefire so shared-module tests run on JDK 17.
+                     The default (2.12.4) crashes on JDK 9+ (broken forked classpath). -->
```
notshivansh, line 151: does this mean, we have two surefire plugins ?

apps/database-abstractor/Dockerfile
```diff
@@ -1,7 +1,19 @@
+
 FROM jetty:9.4-jre8
 USER root
+# Swap the baked-in Java 8 for a maintained Java 17 JRE at the same path.
```
notshivansh, line 4: no swap in.
use expected jre directly.

pom.xml
ayushaga14: there's only 1 declaration

apps/database-abstractor/Dockerfile
ayushaga14: can do this when we later love to jetty 12

notshivansh (approved):

PR #5941, opened 2026-07-31
Fix violations type filtering

notlazykid:


apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/guardrails/violations/ViolationsPage.jsx
```diff
@@ -338,7 +354,7 @@ function transformEvent(event, collectionsMap, usernameMap) {
         id: event.id,
         apiCollectionId: event.apiCollectionId,
         detected: event.timestamp,
-        type: typeLabel,
+        type: classifyPolicyType(policyName),
```
ayushaga14, line 357: why did we change type key value?

notlazykid: Type column used request-shape but the pie uses policy name, so they didn't match and filtering broke. Switched type to classifyPolicyType(policyName) to align them. Kept typeLabel since agenticAssetTag and evidence extraction still need it.

rohan-akto (approved):

PR #5945, opened 2026-08-01
miniruntime: add agent graphs to copilot studio multi env

rohan-akto:


libs/dao/src/main/java/com/akto/util/Constants.java
```diff
@@ -97,6 +98,7 @@ private Constants() {}
     public static final String AKTO_COPILOT_BOT_SCHEMA_TAG = "bot-schemaname";
     public static final String AKTO_COPILOT_BOT_ENVIRONMENT_TAG = "bot-environment-id";
     public static final String AKTO_COPILOT_CONVERSATION_URL_PREFIX = "/copilot/conversation";
+    public static final String AKTO_COPILOT_INVENTORY_TAG = "inventory";
```
nayanakto: use copilot-inventory

nayanakto (approved):

PR #5954, opened 2026-08-03
feat : return copilot hostname as received directly

harshithb3304:


apps/mini-runtime/src/main/java/com/akto/hybrid_parsers/HttpCallParser.java
```diff
@@ -370,6 +370,13 @@ public static String getHostnameForCollection(HttpResponseParams responseParam,
                 return baseHostname;
             }
 
+            // COPILOT_STUDIO: copilot-shield sends the final collection name as the raw host already - trust it as-is.
+            String appNameTag = tagsMap.get(Constants.AI_AGENT_APP_NAME);
```
nayanakto, line 374: why do you need this change??
dont pass bot-name tag from the copilot shield..

nayanakto (changes requested):

PR #5956, opened 2026-08-04
Browser Extension settings revamp + config-driven account rows

rahul-akto:
## What

Revamps the **Browser Extension** settings page (`/dashboard/settings/browser-extension`) and makes account-level configs carry the full, config-driven extraction spec the extension consumes.

### Settings page (frontend)
- Deduplicated catalogue, "Configured" hosts list with pagination, per-row actions, and a host picker modal.
- Rich custom-host form (transport / format / prompt path / operations / frameMatch / response + model paths) with a worked example, built on the product's Polaris `Dropdown`.
- Version-support note moved into a small ⓘ tooltip beside the page title (via a new `titleMetadata` passthrough on `PageWithMultipleCards`).

### Config-driven account rows (backend)
- **Enable copies the spec**: `setBrowserExtensionConfigActive` now copies the whole catalogue document into the account row, so a toggled-on host becomes a full config-driven entry instead of bare `host + paths`. The copy is generic — everything flows through except account-owned fields (`_id, host, active`, audit) — so new extraction fields need no code change. Added `BrowserExtensionConfigCommonDao.findRawByHost`.
- **Tolerant read**: account rows now hold polymorphic fields (`path`/`modelPath`/`responsePath` can be a String or a List), which crashed the POJO-codec read and broke the page. The account read now uses the tolerant `BrowserExtensionConfigCommon.fromDocument` mapper (same pattern as the catalogue read); that DTO gained `responseFormat`/`responsePath`/`modelPath` so nothing is dropped.
- Dedup-by-host + `_id` sort on the catalogue so seeded top brands lead.

## Notes
- The public v2 read (PII strip) lives in the **database-abstractor** (separate checkout) and is tracked separately.
- Catalogue data is seeded from the extension's `monitoring-configs.js` as a one-time data step.

## Test
- Open `/dashboard/settings/browser-extension`: configured list loads (no duplicate rows), toggle on/off persists, add/edit custom host works, version tooltip renders.
- Toggling a supported host writes the full spec into the account row.

### DIFF
--- a/apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/settings/browser_extension/BrowserExtensionSettings.css
+++ b/apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/settings/browser_extension/BrowserExtensionSettings.css
@@ -2,12 +2,12 @@
    (hover-reveal, row hover, status dot, custom toggle). Uses Polaris tokens so light/dark
    both track the theme. */
 
-/* row spacing/layout comes from Box padding + HorizontalStack gap; only the hover, the divider
-   between rows, and the disabled dim have no Polaris prop equivalent. */
-.bext-row { border-top:1px solid var(--p-color-border-subdued,#eef0f4); transition:background .1s; }
-.bext-row:first-child { border-top:0; }
-.bext-row:hover { background:var(--p-color-bg-surface-hover,#f6f7fa); }
-.bext-row.off { opacity:.6; }
+/* rows are spaced cards (border/radius/padding via Box props); only the hover lift and the
+   disabled dim have no Polaris prop equivalent. */
+.bext-row { transition:background .12s, box-shadow .12s, border-color .12s; }
+.bext-row:hover { background:var(--p-color-bg-surface-hover,#f9fafb); border-color:var(--p-color-border-emphasis,#cfcfe6);
+  box-shadow:0 1px 3px rgba(17,20,28,.07); }
+.bext-row.off { opacity:.55; }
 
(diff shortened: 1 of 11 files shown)


libs/dao/src/main/java/com/akto/dao/BrowserExtensionConfigDao.java
```diff
@@ -20,9 +26,27 @@ public Class<BrowserExtensionConfig> getClassT() {
...
+                .skip(skip).limit(limit).cursor()) {
+            while (cursor.hasNext()) {
+                BrowserExtensionConfigCommon config = BrowserExtensionConfigCommon.fromDocument(cursor.next());
+                // host is mandatory, so a document without it is not usable
+                if (config != null) {
+                    configs.add(config);
+                }
+            }
+        }
+        return configs;
```
notshivansh: use native functions, from sdk.

apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/settings/browser_extension/BrowserExtensionSettings.css
```diff
@@ -0,0 +1,54 @@
+/* Browser Extension settings — scoped styles for behaviours Polaris props can't express
+   (hover-reveal, row hover, status dot, custom toggle). Uses Polaris tokens so light/dark
+   both track the theme. */
```
notshivansh, line 3: why so much custom CSS. ??

notshivansh (dismissed):
NEED TO FIX THE CSS.
PLEASE NO CUSTOM CSS.

notshivansh (approved):

PR #5957, opened 2026-08-04
Add v2 browser-extension config API (rich, typed extraction spec)

rahul-akto:
## What

Adds `fetchBrowserExtensionConfigsV2` to the database-abstractor — the endpoint the browser extension calls to load per-site monitoring configs.

- Serves the account-level configs, now carrying the **full extraction spec** the extension supports: `transport, method, format, path, operations, frameMatch, responseFormat, responsePath(s), responseKeyPath, modelPath, modelHeader, triggerFrame, blockResponseFrames, identity, enforceAt`.
- These fields are **polymorphic** (`path`/`modelPath`/`responsePath` are a String *or* a List), so the read maps raw documents via `BrowserExtensionConfigCommon.fromDocument` instead of the POJO codec. This yields a typed object that exposes **only known config fields** — the account row's `createdBy`/`updatedBy` audit/PII never leaks to this public-facing endpoint.
- `BrowserExtensionConfigCommon` extended to model the remaining extension fields (`identity`, `responseFormat`/`responsePath`/`responsePaths`/`responseKeyPath`, `modelPath`/`modelHeader`, `triggerFrame`, `blockResponseFrames`, `enforceAt`).

## Paired with

Dashboard-side changes (catalogue→account spec copy on enable, tolerant read) in akto-api-security/akto#5956.

## Test

`POST /api/fetchBrowserExtensionConfigsV2` returns each active host with its full spec and **no** `createdBy`/`updatedBy`.

### DIFF
--- a/libs/dao/src/main/java/com/akto/dao/BrowserExtensionConfigDao.java
+++ b/libs/dao/src/main/java/com/akto/dao/BrowserExtensionConfigDao.java
@@ -68,7 +68,11 @@ public class BrowserExtensionConfigDao extends AccountsContextDao<BrowserExtensi
                 continue;
             }
             String key = cc.getHost().trim().toLowerCase();
-            commonHosts.add(key);
+            // record the host even when opted out, so it is never re-added as a "custom" host below,
+            // and never listed twice if the catalogue itself repeats a host
+            if (!commonHosts.add(key)) {
+                continue;
+            }
             Document override = accountByHost.get(key);
             if (override != null && Boolean.FALSE.equals(override.get(BrowserExtensionConfig.ACTIVE))) {
                 continue;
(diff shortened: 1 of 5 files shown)


libs/dao/src/main/java/com/akto/dao/BrowserExtensionConfigDao.java
```diff
@@ -25,6 +31,36 @@ public List<BrowserExtensionConfig> findActiveConfigs() {
...
+        try (MongoCursor<Document> cursor = coll.find().cursor()) {
+            while (cursor.hasNext()) {
+                BrowserExtensionConfigCommon config = BrowserExtensionConfigCommon.fromDocument(cursor.next());
+                if (config != null) {
+                    configs.add(config);
+                }
+            }
+        }
+        return configs;
+    }
```
notshivansh, line 59: use native sdk please ??

rahul-akto: done

notshivansh (approved):

PR #6025, opened 2026-08-10
added config handler

harshithb3304:
## Problem

Mini-runtime only reads config from the container's own environment variables. Changing something like `DEBUG_URLS` or `AKTO_LOG_LEVEL` means editing every instance's deployment and restarting each one by hand — no central place to change a value once for a whole fleet.

## Why

Let select env vars be overridden from the DB (`AccountSettings.runtimeEnvOverrides`, written from the dashboard — see PR #6026), so ops can change them once instead of per-instance.

## Flow (as currently implemented)

```
AccountSettings.runtimeEnvOverrides (Mongo)
        │  polled every 30s (+ once synchronously at boot) by ConfigHandler
        ▼
   ConfigHandler.overrides (in-memory map)
        │
        ▼
   readEnv(name, default): DB override → System.getenv → default
        │
        ▼
   Read at class-load / process start only (Main.java, LoggerMaker.java, Utils.java field init)
        │
        ▼
   New value takes effect only after the next restart of that instance
```

Every heartbeat, mini-runtime also reports its current values for these vars into `ModuleInfo.additionalData.env` — the same mechanism other module types (threat-detection, traffic-collector) already use — so the dashboard's Module Info page can show real per-instance state instead of guessing.

## What's in this PR

- New `ConfigHandler`: polls `AccountSettings.runtimeEnvOverrides`, exposes `readEnv(name, default)`.
- `DEBUG_URLS`/`DEBUG_HOSTS` (`Utils.java`) and `AKTO_LOG_LEVEL`, `MINI_RUNTIME_NAME`, `AKTO_CONFIG_NAME`, `AKTO_KAFKA_BROKER_URL`, `AKTO_KAFKA_GROUP_ID_CONFIG`, `AKTO_KAFKA_MAX_POLL_RECORDS_CONFIG` (`Main.java`, `LoggerMaker.java`) now read through `ConfigHandler` instead of `System.getenv` directly.
- `ModuleInfoWorker` reports the whitelisted env values on every heartbeat.
- Deliberately narrow: only these vars, not a blanket `System.getenv` replacement. `VPC_ID` and `DB_MERGING_MODE` were considered and excluded — too load-bearing to convert without more scrutiny.

Companion PRs (same feature): dashboard side https://github.com/akto-api-security/akto/pull/6026, cyborg side https://github.com/akto-api-security/akto/pull/6018

### DIFF
--- a/apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/Main.java
+++ b/apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/Main.java
@@ -11,6 +11,7 @@ import java.util.regex.Pattern;
 
 import com.akto.DaoInit;
 import com.akto.RuntimeMode;
+import com.akto.config.ConfigHandler;
 import com.akto.billing.UsageMetricUtils;
 import com.akto.dao.*;
 import com.akto.dao.context.Context;
@@ -94,7 +95,7 @@ public class Main {
 
     public static boolean isOnprem = false;
     static long lastLogSyncOffsetMRS;
-    static boolean syncImmediately = false;
+    static boolean syncImmediately = true;
     static boolean fetchAllSTI = true;
     private static final String NON_API_FILTER_ORG_ID = "2651c8f0-cf6b-416a-a304-b8d8adc573d7";
     static boolean isNonApiContentTypeFilterEnabled = false;
@@ -259,9 +260,9 @@ public class Main {
 
 
     public static String customMiniRuntimeServiceName;
-    private static final String podName = System.getenv().getOrDefault("POD_NAME", "");
-    private static final String nodeName = System.getenv().getOrDefault("NODE_NAME", "");
-    private static final String miniRuntimeName = System.getenv().getOrDefault("MINI_RUNTIME_NAME", "");
+    private static final String podName = ConfigHandler.readEnv("POD_NAME", "");
+    private static final String nodeName = ConfigHandler.readEnv("NODE_NAME", "");
+    private static final String miniRuntimeName = ConfigHandler.readEnv("MINI_RUNTIME_NAME", "");
     static {
         if (!miniRuntimeName.isEmpty()) {
             // Highest priority: explicit MINI_RUNTIME_NAME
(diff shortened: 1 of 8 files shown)


apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/Main.java
```diff
@@ -94,7 +95,7 @@ private static void printL(Object o) {
 
     public static boolean isOnprem = false;
     static long lastLogSyncOffsetMRS;
-    static boolean syncImmediately = false;
+    static boolean syncImmediately = true;
```
kural-akto: Do we need to change this?

kural-akto (dismissed):

apps/mini-runtime/src/main/java/com/akto/hybrid_runtime/Main.java
```diff
@@ -90,8 +90,7 @@ private static void printL(Object o) {
             debugPrintCounter--;
             loggerMaker.warn(o.toString());
         }
-    }   
-
+    }
```
gauravakto: revert

libs/dao/src/main/java/com/akto/dto/monitoring/ModuleInfoConstants.java
```diff
@@ -0,0 +1,17 @@
+package com.akto.dto.monitoring;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ModuleInfoConstants {
```
gauravakto, line 6: is this a dto ? if not a dto then put somewhere else in utils

harshithb3304: No its just holding constants , there is an equivalent file of this on master repo .

libs/utils/src/main/java/com/akto/metrics/ModuleInfoWorker.java
```diff
@@ -55,6 +59,19 @@ private void scheduleHeartBeatUpdate() {
         scheduler.scheduleWithFixedDelay(() -> {
             try {
             moduleInfo.setLastHeartbeatReceived(Context.now());
+            Map<String, String> allowedKeys = ModuleInfoConstants.ALLOWED_ENV_KEYS_BY_MODULE.get(_this.moduleType);
+            if (allowedKeys != null) {
+                Map<String, Object> env = new HashMap<>();
+                for (String key : allowedKeys.keySet()) {
+                    String value = ConfigHandler.readEnv(key, null);
```
gauravakto, line 66: should the default value be null?

kural-akto (dismissed):

gauravakto (approved):

PR #6026, opened 2026-08-10
added mini-runtime support to edit envs

harshithb3304:
## Problem

An account can run many mini-runtime instances. Today, changing a mini-runtime env var (log level, Kafka config, debug filters) means editing every instance's deployment config and restarting each one by hand — there's no way to change a value once and have it apply fleet-wide.

## Why

Fill a value once in the dashboard, every mini-runtime instance for that account picks it up — no per-instance deployment edits.

## Flow

```
User edits env vars in Module Info UI (mini-runtime row)
        │
        ▼
POST /api/updateRuntimeEnvOverrides  ──▶  AccountSettings.runtimeEnvOverrides (Mongo, one doc per account)
        │
        ▼
POST /api/rebootModules (active MINI_RUNTIME instances only)
        │
        ▼
ModuleInfo.reboot = true  ──▶  existing heartbeat mechanism picks it up (~30s) ──▶ instance restarts
```

Two read paths for the UI itself:
```
Module Info table  ◀── ModuleInfo.additionalData.env  (real, per-instance values, reported every heartbeat)
Env editor modal   ──▶ writes to AccountSettings.runtimeEnvOverrides (account-wide, not per-instance)
```

## What's in this PR

- New `AccountSettings.runtimeEnvOverrides` field + `api/updateRuntimeEnvOverrides` endpoint (per-key merge, not whole-map replace).
- `MINI_RUNTIME` added to the Module Info page's configurable module types, reusing the existing per-module env editor UI/component as-is.
- `MINI_RUNTIME` added to `rebootModules`'s allowed-module filter — without this, instances named via `MINI_RUNTIME_NAME` (not the `Default_`/`akto-mr` fallback) never actually got rebooted.
- Env editor now only submits changed keys, so an untouched field can't accidentally overwrite a real value with blank.

Companion PRs (same feature): mini-runtime side https://github.com/akto-api-security/akto/pull/6025, cyborg side https://github.com/akto-api-security/akto/pull/6018

## Verified

5 concurrent mini-runtime instances: saved `DEBUG_URLS`/`DEBUG_HOSTS` from this UI, confirmed live pickup within ~30s with no restart; saved `AKTO_LOG_LEVEL`, confirmed all active instances rebooted and the fresh boot reflected the new value.

### DIFF
--- a/apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/settings/module_info/ModuleInfo.jsx
+++ b/apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/settings/module_info/ModuleInfo.jsx
@@ -1,4 +1,4 @@
-import { Button, LegacyCard, DataTable, Checkbox, Modal, Text, Tooltip, Icon, HorizontalStack } from "@shopify/polaris"
+import { Button, LegacyCard, DataTable, Checkbox, Modal, Text, Tooltip, Icon, HorizontalStack, Banner } from "@shopify/polaris"
 import { AlertMinor } from "@shopify/polaris-icons"
 import { useEffect, useState } from "react";
 import settingRequests from "../api";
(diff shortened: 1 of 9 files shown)


apps/dashboard/web/polaris_web/web/src/apps/dashboard/pages/settings/module_info/ModuleInfo.jsx
```diff
@@ -70,8 +73,19 @@ const ModuleInfo = () => {
 
     const handleSaveEnv = async (moduleId, moduleName, envData) => {
         try {
-            await settingRequests.updateModuleEnvAndReboot(moduleId, moduleName, envData);
-            func.setToast(true, false, "Environment config saved successfully. Module will reboot.");
+            if (selectedModule?.moduleType === 'MINI_RUNTIME') {
```
kural-akto, line 73: Why are we doing this, instead of updating in moduleInfo?

kural-akto (dismissed):

gauravakto (approved):

PR #6078, opened 2026-08-13
Fix/remove init logs

harshithb3304:


libs/utils/src/main/java/com/akto/runtime/utils/Utils.java
gauravakto: return ret ??? why empty hasheset created again?

gauravakto (approved):

libs/utils/src/main/java/com/akto/runtime/utils/Utils.java
```diff
@@ -226,16 +226,11 @@ public static String printDebugHostLog(HttpResponseParams httpResponseParams) {
...
         String debugUrls = ConfigHandler.readEnv("DEBUG_URLS", null);
         if (debugUrls == null || debugUrls.isEmpty()) {
-            ret = new HashSet<>();
-        } else {
-            ret = new HashSet<>(Arrays.asList(debugUrls.split(",")));
+            return new HashSet<>();
         }
-        logger.warn("DEBUG_URLS initialized with: " + ret.toString());
-        return ret;
+        return new HashSet<>(Arrays.asList(debugUrls.split(",")));
```
gauravakto, line 233: actualy wait instead of doing new create a static hashset? 

and before using/return do set.clear?

PR #6122, opened 2026-08-17
feat: add custom redaction support in cyborg

abhijeet-akto:


libs/dao/src/main/java/com/akto/dto/GuardrailPolicies.java
```diff
@@ -61,6 +61,8 @@ public class GuardrailPolicies {
     private Map<String, Object> contentFiltering;
 
     private LLMRule llmRule;
+    
+    private List<RedactionRule> redactionRules;
```
krngrover94, line 65: Do we need to add this in mcp endpoint shield Go library as well?

krngrover94 (approved):

libs/dao/src/main/java/com/akto/dto/GuardrailPolicies.java
abhijeet-akto: Yes added already

PR #6147, opened 2026-08-19
Removing onboarding & other UI changes

notlazykid:


notshivansh (approved):
merging, but might need to revert.

PR #6176, opened 2026-08-21
feat: add threat-client-for-guardrails

abhijeet-akto:


apps/threat-detection/src/main/java/com/akto/threat/detection/tasks/AbstractKafkaConsumerTask.java
```diff
@@ -50,7 +50,7 @@ public void run() {
               logRecordsPerMin(records.count());
               processRecords(records);
 
-              if (!records.isEmpty()) {
+              if (!records.isEmpty() && shouldCommitAfterProcessing()) {
```
ayushaga14, line 53: can this impact existing consumers using this

ayushaga14 (commented):
added comments

apps/threat-detection/src/main/java/com/akto/threat/detection/tasks/AbstractKafkaConsumerTask.java
abhijeet-akto: No it doesn't, default is true which is current behaviour. Guardrails consumer overrides it with false.

ayushaga14 (approved):

PR #6256, opened 2026-08-27
Fix: Other type in Violations by Type graph

notlazykid:


apps/threat-detection-backend/src/main/java/com/akto/threat/backend/service/ThreatApiService.java
```diff
@@ -186,15 +186,19 @@ private ThreatCategoryWiseCountResponse computeSubCategoryWiseCount(
       pipeline.add(new Document("$group",
           new Document("_id", dedupeKey)
               .append("category", new Document("$first", "$category"))
-              .append("subCategory", new Document("$first", "$subCategory"))));
+              .append("subCategory", new Document("$first", "$subCategory"))
+              .append("latestApiEndpoint", new Document("$first", "$latestApiEndpoint"))));
```
ayushaga14, line 190: why do we need an endpoint filter for computing subcategorywise count?

PR #6285, opened 2026-09-01
Add consolidated bulkRecordTestingRunResults API to database-abstractor

gauravakto:
## Why

Investigated a customer report of test runs taking hours. Traced part of the slowdown to how mini-testing persists each finished test result: for every single `(API, test)` outcome, it makes up to **5-7 separate HTTP calls** to database-abstractor (insert result, update run-summary counts, look up existing issues, write issue updates, update issue counts, plus conditional calls for reruns/custom templates) — each its own network round trip, each individually exposed to retries/backoff if the backend hiccups. Profiling showed this persistence tail was 67-76% of the measured per-test cost, well above actual test-execution time.

## What

Adds one new endpoint, `bulkRecordTestingRunResults`, that does all of that work for a batch of results in a single call. It's a straight port of the existing per-result logic (same business rules, same Mongo documents, same semantics) — not a redesign — just collapsed from N round trips into 1.

This PR is intentionally scoped to **just the new endpoint**: nothing existing is modified. No existing route, action method, or `DataActor`/`DbLayer` method changed — this is purely additive, so it can't regress anything currently in production. The mini-testing side (actually calling this instead of the old per-result flow) is a separate, follow-up change — this PR only adds the capability.

## Test plan

- [x] Compiles clean (`libs/dao`, `libs/utils`, `apps/database-abstractor`).
- [x] Isolated benchmark (new endpoint vs. the equivalent old per-result calls, same payload, same local Mongo): consolidating to one call was ~2x faster even on localhost with no real network latency — expected to be a bigger win in production where each old round trip pays real WAN/hybrid latency.
- [x] Manually verified via the local benchmark harness that a passing/first-seen result persists correctly through the new endpoint.
- [ ] Deploy to a shared non-prod environment (doesn't touch existing APIs, so safe to deploy ahead of any caller using it) and run mini-testing locally against it with the caller-side change, to verify the less-common paths end-to-end:
  - a genuinely new vulnerability creates its issue document (this path had a real bug, since fixed - see commit history)
  - an existing issue's `collectionIds` survive a repeat update (also previously buggy, since fixed)
  - run-summary counts (`testResultsCount`, `countIssues`) come out correct after a mixed batch of passing/vulnerable results
  - a rerun-replace (`rerunDeleteIds`) correctly deletes the prior result


gauravakto:
Follow-up fix pushed: the Kafka secondary-write fallback (\`kafkaUtils.insertDataSecondary(...)\` in the catch blocks of both \`insertTestingRunResults\` and the new \`bulkWriteTestingRunResults\`) ships raw \`TestingRunResult\` payloads, not the \`List<BulkUpdates>\` shape every other bulk-write trigger method uses. \`KafkaUtils.parseAndTriggerWrites\` unconditionally parsed the payload as \`List<BulkUpdates>\` before dispatching, and neither trigger method had a registered \`case\` — so on replay the parse would throw, or the write would silently fall through to \`default: break;\` and be dropped.

Added early-return branches for both trigger methods (same pattern already used for \`insertEndpointShieldLog\`/\`storeAgentQueryRecords\`), which drive the raw payload through \`DbAction.insertTestingRunResults()\`/\`bulkWriteTestingRunResults()\` directly so the same conversion logic (workflow node remap, aiSummaryTraces, id conversions) runs as on the live HTTP path.

Note: the \`insertTestingRunResults\` gap predates this PR — fixed it alongside since both needed the identical fix.

(Today this secondary topic isn't actually consumed by anything in this repo — \`initKafkaConsumer()\` only subscribes to the primary topic — so this was latent, not actively dropping data. Fixing it now so it's correct if/when that changes.)

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -2808,55 +2737,63 @@ public String insertTestingRunResultSummary() {
...
-                            ((Map) x.get(tmp)).remove("authMechanism");
-                            ((Map) x.get(tmp)).remove("customAuthTypes");
-                            data.put(tmp, objectMapper.convertValue(x.get(tmp), YamlNodeDetails.class));
-                        }
+    /**
+     * Converts a raw TestingRunResult payload (as received over HTTP) into a {@link TestingRunResult},
+     * handling the workflow node details remap, aiSummaryTraces placement, testResults normalization
+     * and hex-id -> ObjectId conversions. Shared by the single-insert and bulk-insert endpoints.
+     */
+    private TestingRunResult buildTestingRunResultFromPayload(BasicDBObject rawPayload) throws Exception {
```
gauravakto, line 2745: this is just extraction into a separate method.

apps/database-abstractor/src/main/java/com/akto/action/DbAction.java
```diff
@@ -2808,55 +2737,63 @@ public String insertTestingRunResultSummary() {
...
+            ObjectId id = new ObjectId(testingRunResult.getTestRunResultSummaryHexId());
+            testingRunResult.setTestRunResultSummaryId(id);
+        }
+
+        return testingRunResult;
+    }
 
+    public String insertTestingRunResults() {
+        try {
+            TestingRunResult testingRunResult = buildTestingRunResultFromPayload(this.testingRunResult);
```
gauravakto, line 2796: can this cause bugs? struts null etc.

Ark2307 (approved):

PR #6292, opened 2026-09-01
fix: table pagination

notlazykid:


notlazykid:
1. **What changed:** every table now has a "Show 20/50/100" dropdown, defaulting to 50, set up centrally, so no per-page edits needed.
2. **Bug:** Selecting a larger page size (e.g. 100) grows the whole page's scroll instead of scrolling inside the table.
3. **Cause:** Polaris IndexTable's sticky header tracks only the page's global scroll, not a scrollbox inside it, boxing the table breaks the header, so it wasn't done and that is big change within itself.

notshivansh (approved):

PR #6343, opened 2026-09-08
add: fetch mutliple agents

notlazykid:


notlazykid:
1. `AgenticObserveAction.java`: added one API that can look up many agents at once (instead of one-by-one), told it to also say which of an agent's connections are LLMs, and made each agent row carry an ID the frontend can look it up by.
2. `struts.xml`: registered that new API so the app can actually call it (plus the one-line fix that made its data arrive in the right shape).
3. `api.js`: the frontend function that calls that new API and reads the result.
4. `DeviceFlyout.jsx`: the endpoints device panel: fetches each agent's connections and tools, and lays the graph out so every agent sits next to its own stuff without things overlapping.
5. `AssetTopologyGraph.jsx`: the shared graph drawing: gave Tools their own colour/icon so they don't look like MCP servers, and let the box grow taller when there's more to show.

notshivansh (dismissed):

notlazykid:
1. Graph height: I'd said earlier the box grows taller when there's more to show. That's changed: it's now a fixed size, the same as the graph on the Agentic Assets page, so both screens look consistent. You pan/zoom for anything that doesn't fit.
2. Where the graph opens: it used to zoom out to fit everything, which made a busy device unreadable. It now opens centred on the thing you actually clicked: the user on the Endpoints page, the agent (or MCP server, or whichever asset type) on Agentic Assets. It re-centres as the data finishes loading.
3. Agentic Assets page now shows an agent's MCP servers → their tools too, the same extra hop the Endpoints page shows. This never worked before: the MCP boxes were missing the ID needed to look their tools up, so the code that was meant to fetch them never ran.
4. Removed duplicated code: the "hang tools off an MCP" logic had been written twice, once per screen, and the two copies had already started behaving differently. It's now one shared file (topologyTools.js) that both screens use, so they can't drift again. About 56 lines lighter overall.
5. Cut a burst of requests: opening a device panel used to fire roughly 90 network calls to load the agents' built-in tools (one set per agent per collection). It's now 3 for the whole panel, using the batching helper that already existed.

notshivansh (approved):

PR #6348, opened 2026-09-08
context graph fix

notlazykid:


notlazykid:
Fixed the "tool tied to itself" bug (agent showing up as its own MCP server) and made the endpoints list smarter about telling agents apart from servers.

rakshak-akto (approved):

PR #6390, opened 2026-09-11
Feature/load secrets from file

ayushaga14:


libs/utils/src/main/java/com/akto/PayloadEncodeUtil.java
```diff
@@ -25,9 +26,9 @@ public class PayloadEncodeUtil {
 
     public static RSAPrivateKey getPrivateKey() {
         try {
-            String privateKeyPem = System.getenv("PRIVATE_KEY");
+            String privateKeyPem = SecretUtils.readSecret("PRIVATE_KEY");
```
gauravakto, line 29: read all these at class load at one place? maybe in static block?

gauravakto (approved):
lgtm, do regression testing

